Issue #13, opened 2016-08-30
[Repo, Build] Move all SCSS to top-level scss/ folder, have build step copy it in


yozlet:
Part of overall `assets` cleanup. (#8)

At the end of this we should be able to remove the `assets/css/` folder from the repo, _but_ we'll need to check that the `scss` build step creates it and the `vendor` subfolder (for #12) when copying files over.

Issue #12, opened 2016-08-30
[Repo, Build] Remove prism.css from repo


yozlet:
Part of overall `assets` cleanup. (#8)

`git rm assets/css/vendor/prism.css` ~~then have it copied there by the `gulp sass` task~~ (no longer needed, see below).

yozlet:
It appears that all the Prism styles are now defined in `styleguide.scss` so the `prism.css` file isn't needed any more.

Issue #8, opened 2016-08-25
[Repo] Remove assets folder from the repo, only have it created by the build process


yozlet:
## Description

Everything in the `assets/` folder should be there as a result of the build process, not because it's been committed there. All assets that belong to the docs (as opposed to those from a vendor/module, including the `uswds` package) should start in a separate folder (e.g. `src/`).

- [x] Remove images
  - [x] Remove all USWDS package images (unfortunately, they're mixed in with the docs images) (#9)
  - [x] Move all docs images to `src/images/`, have build step copy them in (#10)
  - [ ] ~~Ideally, images copied from `uswds` should go to `assets/img/vendor/uswds/` (but this will require lots of doc changes - probably automatable. We may want to postpone this work to a later date.)~~ (Postponed, no longer part of this issue, but still in consideration at #11)
- [x] Remove styles
  - [x] Make sass build step copy `prism.css` from `node_modules/` (#12)
  - [x] Move all docs SCSS to `src/scss/`, have build step copy them in (#13)
- [x] Remove Javascript
  - [x] Make JS build step copy vendor JS from `node_modules/` (#14)
  - [x] Move all docs JS to `src/js/`, have build step target `assets/js/` for `styleguide.js` output (#15)
- [x] Remove `assets/` (#16)

rogeruiz:
This shouldn't be entirely necessary anymore @yozlet. PrismJS has been removed since it is now longer needed to generate `<markup>` code on the client. It gets built in `config/gulp/html.js` now.

yozlet:
@rogeruiz Interesting, somehow `config/gulp/html.js` didn't make it over to this new repo. `website.js` also seems to be missing, yet the site builds OK. (I probably don't fully understand what `website.js` is for.) Should I just copy them over? @juliaelman?

juliaelman:
@yozlet good questions and I have some answers! 🎉  Going to write this up here as a reference point on how the gulp tasks have been updated/changed from the original repo to this one.

`html.js` needs to be added in and will [be part of this work](https://github.com/18F/web-design-standards-docs/issues/6). These tasks are not in this repo b/c the addition of that `gulp` task was added *after* the initial implementation of this repo. Check out [this PR](https://github.com/18F/web-design-standards/pull/1370) for that implementation.

[...]

PR #18, opened 2016-09-02
[Repo, Build] Move css assets

yozlet:
Implements #12 and #13 as part of #8. Note that instead of being copied in, `prism.css` is just removed.


juliaelman:
LGTM! 🎉

Issue #15, opened 2016-08-30
[Repo, Build] Move all custom JS to top-level js folder


yozlet:
Part of overall `assets` cleanup. (#8)

Also tweak `config/gulp/javascript.js`: change

    entries: 'assets/js/start.js'

to 

    entries: 'js/start.js',

Issue #14, opened 2016-08-30
[Repo, Build] Remove vendor JS, have it copied from node_modules by build step


yozlet:
Part of overall `assets` cleanup. (#8)

At present the only vendor JS needed is jQuery. ~~(Prism can just go, as the syntax highlighting work is now done during the build, in `config/gulp/html.js`.)~~ (Prism's still needed until #6 is complete)

PR #19, opened 2016-09-02
[Repo, Build] Move js assets

yozlet:
Implements #14 and #15 as part of #8. Nearly done!


config/gulp/javascript.js
```diff
@@ -27,6 +27,42 @@ gulp.task('eslint', function (done) {
...
+  dutil.logMessage(task, 'Copying JS from js/');
+
+  var stream = gulp.src('./js/**/*')
+    .pipe(gulp.dest('assets/js/'));
+
+  return stream;
+
+});
+
+gulp.task('copy-jquery-javascript', function (done) {
```
juliaelman: Does this need to be a task or can we just have it as part of the `package.json` in this repo?

juliaelman:
Just to clarify, are we relying on the `prism.js` file imported via the Standards?

config/gulp/javascript.js
yozlet: What do you mean, as part of `package.json`? (Though, now that you point it out, I realised that we don't actually need to be copying any JS (other than `uswds.js`) anywhere - the build process is going to pull the vendor JS from `node_modules` during the compilation stage, so it doesn't need to be in `assets/js`. Silly me. I'm going to fix that.)

yozlet: Oh, duh, sorry. We're saying the same thing. No, you're right, it doesn't need to be a task because it's already in `package.json` and so Browserify will pull it in anyway.

yozlet:
Nope, it's in this repo's own `package.json`.

juliaelman:
🚀  nice work!

Issue #16, opened 2016-08-30
[Repo, Build] Remove `assets/`, have it created at start of build


yozlet:
Part of overall `assets` cleanup. (#8)

  - [x] Build step creates `assets/` at the start
  - [x] `git rm assets/` and add to `.gitignore`
  - [ ] Ensure that `vendor` subfolders (inside `img`, `css` and `js`) are all being created properly

juliaelman:
@yozlet looks like the last part for this is placing the `uswds` into a vendor folder. Happy to tackle that if you'd like! 😄

Issue #11, opened 2016-08-29
[Repo, Build] Images from USWDS package should go in assets/img/vendor/uswds/


yozlet:
Part of overall `assets` cleanup. (#8)

This would be good for neatness's sake, and to avoid accidental name collisions. However, it'll require all existing image references in the docs to be reworked (which may be automatable, or possibly easier to just build without USWDS images and then check Web Inspector to see which images 404ed).

shawnbot:
I've been thinking about this lately as we consider migrating to a full-blown Jekyll theme (#130), and I'm growing increasingly more comfortable with a directory structure under `/assets/uswds/` that has its own `css`, `fonts`, `img`, and `js` subdirectories. I like this because it means that we can configure the USWDS asset directory in one place and template the paths of different asset types in the same way, which makes search-and-replace operations a lot easier. So instead of this:

```html
<img src="{{ site.baseurl }}/assets/img/uswds/icon.svg">
```

which is hard to generalize for different asset types because the `img` bit goes between the base URL and the USWDS "path", we do this:

```html
<img src="{{ site.baseurl }}/assets/uswds/img/icon.svg">
```

which makes it easier to generalize the paths for USWDS assets because they all share a common base. Thoughts, anyone?

yozlet:
Makes sense to me!

juliaelman:
@shawnbot yes, please! Thanks for getting our conversation out in writing.

We should also think about the [html folder](https://github.com/18F/web-design-standards/tree/staging/src/html) in the main repo as part of this work too.

Going to open a new issue for this work and close this one.

juliaelman:
New issue opened here: https://github.com/18F/web-design-standards-docs/issues/166

PR #20, opened 2016-09-02
[Repo, Build] Remove assets folder

yozlet:
Implements #16, finishing #8 (with the exception of #11, which we should deal with separately)

**Note:** This PR follows #19, so merge that in first!


config/gulp/javascript.js
```diff
@@ -27,6 +27,31 @@ gulp.task('eslint', function (done) {
 
 });
 
+gulp.task('copy-jquery-javascript', function (done) {
+
+  dutil.logMessage(task, 'Copying JS from jQuery');
+
+  var stream = gulp.src('./node_modules/jquery/dist/jquery.min.*')
+    .pipe(gulp.dest('assets/js/vendor/jquery/'));
```
juliaelman: Is there a reason this needs to live inside of a folder? Or can the file live in the `vendor` folder on it's own?

config/gulp/javascript.js
```diff
@@ -27,6 +27,31 @@ gulp.task('eslint', function (done) {
...
+
+/* TODO: Remove this build step *and* its invocation in the main task
+   once build-time syntax colouring is restored
+*/
+gulp.task('copy-prismjs-javascript', function (done) {
+
+  dutil.logMessage(task, 'Copying JS from PrismJS');
+
+  var stream = gulp.src('./node_modules/prismjs/prism.js')
+    .pipe(gulp.dest('assets/js/vendor/prismjs/'));
```
juliaelman: Is there a reason this needs to live inside of a folder? Or can the file live in the `vendor` folder on it's own?

juliaelman:
Looks good! Added a few notes inline and let me know if those make sense or not.

Also, I submitted a PR to add the close images to the `img` folder, refs #20

yozlet:
@juliaelman I put them in folders because it felt like a better way to organise it, but I don't have massively persuasive arguments in favour of the change, so I'm happy to leave it how it was if you prefer.

juliaelman:
@yozlet cool! I just wasn't sure if they were being put into folders b/c of reasons 😄  Yeah, let's change that so that files output to the `vendor` directory itself - thank you!

yozlet:
Gah, I'm an idiot. I forgot that I ripped all that vendor-lib-copying code out because Browserify doesn't need it.

PR #23, opened 2016-09-06
Jh readme split update

jameshupp:
## Changes
- included README language about the split
- a few minor stylistic changes
- changed from UL to OL for some instructions

## Review
@juliaelman 
@kategarklavs 

## Notes
No table of contents here since at present this README is so much shorter than the WDS README is and there’s a slightly awkward TOC thing with the use of `uswds` in one of the headers. But we can figure that out later without much trouble.


kategarklavs:
@jameshupp Just reviewed — this looks great!

Issue #24, opened 2016-09-07
Add alert to WDS site to inform users of repo split


kategarklavs:
## Description

As part of the larger repo-split comms plan, I'd like to add an alert to standards.usa.gov to let users know about the repo split. Here's the proposed copy:

`The Standards' GitHub repos have changed! [web-design-standards](https://github.com/18F/web-design-standards) is the code base for the Standards themselves, and [web-design-standards-docs](https://github.com/18F/web-design-standards-docs) contains the code and documentation for this site.`

My thought is that this alert could stay up for several weeks (perhaps up to a month?), just to give folks enough time to get used to the change. 

As far as placement of the alert goes, this is something I need your helps with! I'm thinking it might be good above all of the other page content — what say you, visually inclined folks?

cc @maya @juliaelman @lauraGgit @donjo @jameshupp

maya:
Maybe we can link `web-design-standards` and `web-design-standards-docs` to their repos too.

Also, I think you meant to tag @lauraponce :)

kategarklavs:
@maya Indeed! (I saw Laura G commenting in the wds channel & then accidentally added her!). @lauraponce — apologies for the missing tag 😌

kategarklavs:
@maya — yep! +1 to linking :)

jameshupp:
If we're using one of our alerts, we'll need separate copy for the header and the body. So something like ...

    ### Changes to the Github repos for the Standards
    
    There are now separate repos for the Standards and for the documentation of 
    the Standards. [web-design-standards](https://github.com/18F/web-design-standards) is 
    the code base for the Standards themselves, and [web-design-standards-docs]
    (https://github.com/18F/web-design-standards-docs) contains the code and 
    documentation for this site.

kategarklavs:
Excellent call, @jameshupp! And +1 to the header copy -- my only recommendation would be to capitalize the "H" in GitHub. 

@maya — is there anything else that you all need, content-wise? Or is this good to go?

maya:
@kategarklavs content looks good! 

The next thing we need is guidance on placement from @donjo or @bradnunnally.

kategarklavs:
@maya Exxxxcellent! I'm glad to hear it 😌

donjo:
I mocked this up quick in chrome dev tools. Is this roughly what we were thinking, placement-wise?  Should it show up in every category? or just on the overview? 

[...]

PR #29, opened 2016-09-12
[Website] add repo split alert to overview and developers page

donjo:
## Description

This adds a temporary announcement to the overview and developers page (under Getting Started) to notify users of the repo split. As defined in #24. 

## Additional information

screen shot 2016-09-12 at 12 39 23 pm

Could use feedback on whether I put all this code in the right place given it's temporary nature.


_includes/repo-split.html
```diff
@@ -0,0 +1,7 @@
+<!--temporary alert to announce repo split-->
+<div class="usa-alert usa-alert-info reposplit-alert">
```
maya: Consider renaming class to `repo_split-alert`.

maya:
@donjo when checking this out, the margins look different from your screenshot. This is what I see:

screen shot 2016-09-12 at 12 30 20 pm
screen shot 2016-09-12 at 12 30 31 pm

maya:
I had to re-run `npm start`, this looks good to me!

PR #30, opened 2016-09-12
Remove test deploy config, replace with live version

yozlet:
... now that we can push the repo live! Yay!

REMEMBER, after merging this, `staging` also needs to be merged to `master`.


juliaelman:
Looks great - we will need to add in the `release` deploy as well and then we are GTG! 😄

juliaelman:
LGTM

PR #49, opened 2016-09-22
Update css watch cmd to point to the correct styles dir for uswds.

juliaelman:
The `watch` command currently points to the incorrect directory for the WDS packaged styles, which therefore does not update the styles correctly when developing locally. This PR solves the crime!


maya:
🚀

PR #51, opened 2016-09-26
adding images for page template examples

donjo:
## Description

Adding images for the page template examples.


maya:
@donjo looks pretty good. I think there's too much padding on the left and right - I'd crop it in so it's visually around 30px on each side.

PR #54, opened 2016-10-03
Clean up npm scripts

### LINKED ISSUE #50: npm "watch" scripts should not run "jekyll build"
I was reviewing #46 today and followed the [instructions] for using the latest version of uswds with the docs, but I ran into a problem: running `npm run watch` (for the first time?) triggers several calls to `bundle exec jekyll build`, and I _think_ that one or more of these failed for me because several different Jekyll processes were trying to build the site in the same directory at the same time. Whatever the cause, I couldn't run the server (`npm start`) and watch files (`npm run watch`) at the same time.

I _think_ that the solution for this is to remove the `&& bundle exec jekyll build` commands from the [`build-*` scripts][build scripts]. Whatever process publishes the site should be using the `build` (instead of, say, `build-css` and `build-js`), and if you're running the Jekyll server the new builds should be incorporated into the local site by Jekyll anyway.

[instructions]: https://github.com/18F/web-design-standards-docs#using-the-latest-version-of-the-uswds-package
[build scripts]: https://github.com/18F/web-design-standards-docs/blob/staging/package.json#L8-L12

shawnbot:
This fixes #50 and tidies up a couple of the other scripts to simplify the install + build + watch process:

- [x] Remove the `&& bundle exec jekyll build` step from each of the asset-specific `build-*` scripts.
- [x] The `build` and `prestart` scripts no longer call `bundle`. Instead, `bundle` is called once in the `postinstall` hook (which is called at the end of a successful `npm install`).
- [x] Nix the `npm run build` step from the README.
- [x] Fix a linting error in a [re-declared variable](files#diff-bd172c7b464c275bb7733e999c156f55R18).

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,12 +29,6 @@ To start, run the following command to install project dependencies:
 npm install
 ```
 
-Next, you’ll perform a build step to copy and compile all of the necessary files that create the site:
-
-```sh
-npm run build
-```
-
 Now that all of your dependencies are installed, you can run your local server by running the following command:
 
 ```sh
--- a/_includes/download-buttons-design.html
+++ b/_includes/download-buttons-design.html
@@ -3,8 +3,8 @@ These `uswds-assets-*.zip` files are generated in 18F/web-design-standards-asset
 repository. See: https://github.com/18F/web-design-standards-assets
 {% endcomment %}
 <div class="link_group-download">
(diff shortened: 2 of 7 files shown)


juliaelman:
@shawnbot oh Circle CI and your helpful caching! I rebuilt without the cache after the removal of the  reference to the `uswds` Github repo branch. We are seeing the correct error now 👍

PR #66, opened 2016-10-21
Fix un-hyperlinked URL in maturity scale page

toolness:
The following paragraph is on the [component maturity scale](https://standards.usa.gov/about-our-work/component-maturity-scale/) page:

> Our system is inspired by http://usajobs.github.io/design-system/getting-started/#maturity, who outline a maturity scale of their own in their Getting Started and who use these labels on the individual component page.

While GitHub is hyperlinking that URL above, it is *not* hyperlinked on the generated site, which requires users to manually copy and paste it into a separate browser tab.  However, it's also difficult to read as a raw URL, so I made it hyperlinked text and simplified the sentence a bit:

> Our system is inspired by the [USAJOBS Design System maturity scale](http://usajobs.github.io/design-system/getting-started/#maturity), which uses similar labels on its individual component pages.


juliaelman:
Nice catch, thanks @toolness!

PR #71, opened 2016-11-09
Removed the line stating jQuery 2.2.0 is included with the USWDS. jQu…

teebolt:
## Description

The documentation currently suggests that uswds.js includes jQuery 2.2.0. This can cause issues and break pages if a.) there is an expectation that jQuery will be loaded to the global namespace or b.) another version of jQuery is already being loaded.

## Additional information

I am not sure what the plan is for uswds.js. There seems to be some relics of when jQuery was actually bundled. 

I would recommend not having any jQuery or reference to jQuery unless jQuery is required as a dependancy.

### DIFF
--- a/pages/getting-started/developers.md
+++ b/pages/getting-started/developers.md
@@ -55,10 +55,6 @@ of your downloaded assets. And the un-minified version is better if you are in a
 development environment or would like to debug the CSS or JavaScript assets in
 the browser. The examples above recommend using the minified versions.
 
-This version of the Standards includes jQuery version `2.2.0` bundled within the
-JavaScript file. Please make sure that you're not including any other version
-of jQuery on your page.
-
 And that’s it — you should be set to use the Standards.
 
 ### Using npm
--- a/pages/getting-started/developers.md
+++ b/pages/getting-started/developers.md
@@ -59,6 +59,8 @@ And that’s it — you should be set to use the Standards.
 
 ### Using npm
 
+Note: Using npm to install the Standards will include jQuery version `2.2.0`. Please make sure that you're not including any other version of jQuery on your page.
+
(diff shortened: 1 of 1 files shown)


juliaelman:
@teebolt thank you for your pull request! 🎉  We've added this disclaimer in our "Getting Started" docs, as we do [include jQuery 2.2.0](https://github.com/18F/web-design-standards/blob/staging/package.json#L39) bundled into our latest version of the Standards.

However, the team has been working towards removing jQuery completely so as not to cause any conflicts in versioning or performance. You are welcome to follow along [here](https://github.com/18F/web-design-standards/issues/1351) where the last bit of the removal is happening.

Closing for now, but we will be sure to remove this disclaimer once we are no longer bundling jQuery.

teebolt:
That may be if someone is using npm to install the standards but downloading the code directly doesn't include jQuery.

juliaelman:
@teebolt indeed! What do you think about us changing the note to state that this is for the `npm` installation use case rather than making an overarching statement? Maybe change to:

> This version **distributed npm package** of the Standards includes jQuery version `2.2.0` bundled within the JavaScript file. Please make sure that you're not including any other version of jQuery on your page.

teebolt:
Works for me 👍

juliaelman:
Awesome! I'll re-open this if you want to make that change so we can get your contribution in here 😄 

I'll be on the lookout to review/merge!

teebolt:
Actually I'm wondering if it will be clear because it still says "bundled within the JavaScript file." Maybe it should go under the **Using npm** section? 

If someone installs via npm, does it actually bundle js code into uswds.js or does it include it as a dependency?

juliaelman:
@teebolt yes! I like this even better 👍

juliaelman:
Looks great and thank you for your contribution @teebolt! 🚀

PR #75, opened 2016-11-28
Updated the instruction to be clear that npm commands need to be ran at the folder level in terminal.

bradnunnally:
Added some instructions that clarified that npm commands need to occur at the folder in terminal.


juliaelman:
✨

PR #82, opened 2016-12-06
Add Politespace

shawnbot:
This is the companion PR to 18F/web-design-standards#1606 that adds [Politespace](https://github.com/filamentgroup/politespace) to the docs repo so that we can demo its input masking in the [address form template](https://standards.usa.gov/form-templates/#address-form). I've also taken the liberty of better organizing the various components so that they each live in their own files.

Other updates:
- Replace the duplicated Politespace jQuery plugin with one from their source directory
- Add `.jekyll-metadata` to `.gitignore`


js/components/sidenav.js
```diff
@@ -0,0 +1,56 @@
...
+  var hashLocation  = $(this).attr('href').split('#')[ 1 ];
+  var scrollTopPos  = calculateAnchorPosition(hashLocation);
+
+  //if anchor doesn't exist on the page, or calc fails
+  //then exit gracefully
+  if (scrollTopPos === 0) {
+    return true;
+  }
+
+  e.preventDefault();
```
msecret: Should this go at the top?

js/components/sidenav.js
```diff
@@ -0,0 +1,56 @@
...
+$('.sidenav').on('keydown', 'a', function (e) {
+  var ENTER = 13;
+  if (e.which === ENTER) {
+    $(this).data('keypress', true);
+  }
+});
+
+$('.sidenav').on('click', 'a', function (e) {
+  // long url splitting
+  var hashLocation  = $(this).attr('href').split('#')[ 1 ];
```
msecret: Whats the WDS JS style guide? Is applying extra spaces to the left of `=` ok?

shawnbot: I'm not sure this is a linting error, but I'll remove it.

js/components/sidenav.js
shawnbot: Honestly, I'm not sure. This part is unchanged, though.

js/components/sidenav.js
shawnbot: Fixed.

PR #87, opened 2016-12-09
Add html to use Politespace on address form example

juliaelman:
NOTE: This pull request is slated to merge into [the work @shawnbot has started](https://github.com/18F/web-design-standards-docs/pull/82) around isolating Politespace to the website only. The goal is to keep the automation work @rogeruiz developed for creating code samples, but also makes it so that we keep our example of implementing Politespace on our address form example.

:thinking: should we include a link to the JS file on this site to that makes Politespace work? I think a link to the code on Github may suffice.

### DIFF
--- a/_includes/code/examples/address-form.html
+++ b/_includes/code/examples/address-form.html
@@ -1,0 +1,80 @@
+
+  <form class="usa-form-large">
+    <fieldset>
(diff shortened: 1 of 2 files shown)


juliaelman:
Refs https://github.com/18F/web-design-standards/pull/1606

PR #89, opened 2016-12-09
Update research strategy

bradnunnally:
## Description

Include the research strategy into the site that includes some additional services we can provide to partner agencies.


bradnunnally:
This is ready to be merged.

PR #93, opened 2016-12-16
Fix issue on about us side navigation

### LINKED ISSUE #92: Sidenav bug when looking at new research pages
After testing on staging it looks like there is an issue around the side navigation when looking at the new research pages, see screenshot below.

screen shot 2016-12-16 at 11 36 14 am

My initial guess, without doing any digging, is that it is because `/research/` somehow collides with any url pattern that is `/research-*/`.

juliaelman:
Fixes #92 

@bradnunnally does this look okay to you? I changed the verbiage in the sidenav to match "Current research" in place of "Research".

### DIFF
--- a/_includes/sidenav/about-our-work.html
+++ b/_includes/sidenav/about-our-work.html
@@ -15,8 +15,8 @@
       <a {% if page.url contains 'about-our-work/research-strategy' %}class="usa-current"{% endif %} href="{{ site.baseurl }}/about-our-work/research-strategy/">Research strategy</a>
     </li>
     <li>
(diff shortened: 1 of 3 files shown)


juliaelman:
@bradnunnally here is a screenshot too! 😄 

screen shot 2016-12-16 at 11 40 25 am

juliaelman:
@bradnunnally we went ahead and launched to production (my bad!). I'll circle back with you on Monday to see if you have any changes to the updated verbiage on these pages.

bradnunnally:
Np. LGTM

PR #97, opened 2016-12-22
Adds max width overrides that were removed

msecret:
The max-width overrides were removed in the main WDS code, but we
want to keep them here while the team doesn't have a full visual
designer.

Related to [18F/web-design-standards#ms-removing_max_width](https://github.com/18F/web-design-standards/tree/ms-removing_max_width), but
doesn't require it to be merged.

### DIFF
--- a/css/styleguide.scss
+++ b/css/styleguide.scss
@@ -8,6 +8,18 @@
 
 //  -------------- //
 
+.usa-content p:not(.usa-font-lead) {
+  max-width: $text-max-width;
+}
+
+.usa-content-list {
+  max-width: $text-max-width;
+}
+
+.usa-font-lead {
+  max-width: $lead-max-width;
+}
+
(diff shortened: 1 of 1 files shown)


juliaelman:
@msecret we probably need to somehow single out the page templates and possibly other demos to reflect this change too `max-width` change too. 🤔  Right now, the page templates still carry over that `max-width` value set in the `styleguide.scss` file. This could be confusing for folks when implementing and seeing something other than what they are expecting.

msecret:
Yea that sounds right @juliaelman . That was a concern of mine of adding that rule to `styleguide.scss`.

juliaelman:
See https://github.com/18F/web-design-standards/issues/1603#issuecomment-274544954

Closing this issue and taking another approach that is more direct for Federalist.

PR #98, opened 2016-12-23
A11y page

shawnbot:
Fixes 18F/web-design-standards#883.

- [x] Adds an "Accessibility" section to the UI components sidenav
- [x] Adds an "Accessibility" page with a single section for "Showing and hiding content", which describes the suggested mechanisms for showing and hiding content for different users


Gemfile
```diff
@@ -1,9 +1,7 @@
 source 'https://rubygems.org'
 
 gem 'jekyll'
-
 gem 'json'
-
+gem 'redcarpet'
```
shawnbot: Redcarpet is better than kramdown because it supports markup within table cells.

_components/a11y.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Hiding and showing content
+
+There will inevitably be cases in which it you need to selectively "show" and
+"hide" content by either removing it from the visual flow or preventing screen
+readers from reading it. We have settled on the following conventions for each
+use case:
+
+{% capture show %}**show**{% endcapture %}
+{% capture hide %}_hide_{% endcapture %}
```
shawnbot: These variables make it easier to change the table below if we decide that some other word or indicator works better.

_config.yml
```diff
@@ -33,7 +33,11 @@ collections:
     permalink: /:path/
 
 # Build settings
-markdown: kramdown
+markdown: redcarpet
+redcarpet:
+  extensions:
+    - tables
```
shawnbot: Again: redcarpet allows markdown in table cells; kramdown doesn't. The `table` extension isn't enabled by default, though.

_includes/sidenav/ui-components.html
```diff
@@ -1,4 +1,4 @@
-{% assign current = page.url | downcase | split: '/' %}
+{% assign current = page.url | split: '/' %}
```
shawnbot: All of our URLs are lowercase, so there's no need to downcase here.

ducklet:
@shawnbot about the first row in the table re: showing to 'visual browsers' but hiding from screen readers - you're linking to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden), which says "It is similarly incorrect to use this attribute to hide content just from one presentation — if something is marked hidden, it is hidden from all presentations, including, for instance, screen readers."  Maybe this row should be removed, and/or could you please clarify what you are suggesting be accomplished by using the hidden attribute?  Thanks!  cc @edwelker

_components/a11y.md
```diff
@@ -0,0 +1,28 @@
...
+permalink: /accessibility/
+title: Accessibility
+category: UI components
+lead: Accessibility tips to help you deliver delightful experiences to all users.
+order: 25
+---
+
+## Hiding and showing content
+
+There will inevitably be cases in which it you need to selectively "show" and
```
juliaelman: Change to:

> There will be cases in which you need to...

bradnunnally:
This being part of the side nav makes the most sense for right now. Later, when we have more content and strategy around accessibility there is an argument for bringing it up to the main nav though. It would just be part of a larger restructuring of the site. 

Current state LGTM!

juliaelman:
🚀  🎸

PR #106, opened 2017-01-31
[WIP] Add blog link

shawnbot:
This adds a blog link (href: https://18f.gsa.gov/tags/web-design-standards/) to the main nav to fix 18F/web-design-standards#1676. It looks like this:

image

I'm not sure if we want that external link icon here, but it doesn't bother me _that_ much. Thoughts on the actual link text ("What’s new") or the icon, @bradnunnally @juliaelman?


bradnunnally:
@shawnbot There's going to be other things under the What's New page of the main nav, so it won't link off site. Plus, it's generally a bad practice to have a main navigation item link off to another site. In my head, and I'm going to get this down as a wireframe today, I see there being a call to action on the What's New page that directs people to the blog.

PR #120, opened 2017-02-13
Updating footer with new layout for 1.0

juliaelman:
This pull request a few changes to the overall interaction of the navigation scrolling:

- Replaced the our own implementation for a sticky nav with [stickyfill](https://github.com/wilddeer/stickyfill) -- thanks for the suggestion, Shawn! This was due to the fact that the footer is no longer a part of the main content container and has it's own state separate from that now. 
- The "What's New" page has not been created yet and have placed a commented out link that to page for now. Just wanting to make sure that it is in there, but that we can also always keep our code base in a shippable state.
- I changed the file structure for the JS and removed the `component` folder. This is legacy and based on when this repo and the [main WDS repo](https://github.com/18F/web-design-standards) were combined. We now have a `vendor` folder for our third party JS files.


_layouts/styleguide.html
```diff
@@ -9,8 +9,7 @@
   <body class="page-{{ page.title | slugify }}">
 
   	{% include navbar.html %}
-
-    <aside class="sidenav">
+    <aside class="sidenav sticky" style="top: 5rem; margin-bottom: 457px;">
```
shawnbot: Could these inline styles go into the stylesheet with some comments explaining where they come from?

css/styleguide.scss
```diff
@@ -8,18 +8,21 @@
 
 //  -------------- //
 
+.sticky {
+  position: -webkit-sticky;
+  position: sticky;
+  top: 0;
```
shawnbot: We need a newline after this line.

css/styleguide.scss
```diff
@@ -177,12 +175,7 @@
     display: block;
     height: $height-nav-secondary;
     padding-left: 3rem;
-  }
-
-  &.is-scrolled {
-    position: fixed;
-    top: 0;
-    width: 100%;
+    z-index: 10;
```
shawnbot: We should follow our [front end guidance](https://pages.18f.gov/frontend/#css-variables) to "name all used z-indexes with a variable."

js/start.js
```diff
@@ -1,7 +1,15 @@
 'use strict';
 
-require('./components/politespace');
-require('./components/scroll-to-top-for-hash');
-require('./components/search');
-require('./components/sidenav');
-require('./components/sticky-nav');
+require('./vendor/politespace');
+require('./vendor/stickyfill.min.js');
```
shawnbot: Not critical for this PR, but I just wanted to note somewhere that we shouldn't have to copy these scripts to `js/vendor` if we're bundling them. We should be able to just:

```js
require('politespace');
require('stickyfill');
```

_layouts/styleguide.html
juliaelman: I based this on the [the example from `stickyfill`](https://github.com/wilddeer/stickyfill/blob/master/test/index.htm#L119). In this instance I think it makes more sense for them to be inline so that it is the ultimate in hierarchy for defining this interactivity. 💪

css/styleguide.scss
juliaelman: Done here 7de0e32

css/styleguide.scss
juliaelman: Done here 527a6f5

juliaelman:
@shawnbot made those changes. Look okay to :shipit:

PR #128, opened 2017-02-14
Fix image path; remove unused images

shawnbot:
This is a fix for a problem introduced by the new `$site-image-path` SCSS variable that I introduced. It makes site images work from `{{ site.baseurl }}/img/` and obviates the need to copy them from `img/` to `assets/img/` in our gulp task. You can see the error in action on the [header demo page](https://standards-staging.usa.gov/headers/extended/), where the left arrow in the back button isn't working:

image

I've also taken the liberty of removing some unused images from our `img` directory.


juliaelman:
@shawnbot looks good, but want to talk a bit about how we are planning to serve compiled assets in general. It seems odd that we have both `img` and `assets/img`.

Merging in, but something we will want to address.

PR #135, opened 2017-02-16
Change Overview to UI components in nav.yml

juliaelman:
Replacing the term `Overview` with `UI components` to be consistent with our other top level nav elements.

### DIFF
--- a/_data/nav.yml
+++ b/_data/nav.yml
@@ -17,7 +17,7 @@ mobile:
   
 
 UI components:
-  - text: Overview
+  - text: UI components
     href: /components/
   - href: /components/typography/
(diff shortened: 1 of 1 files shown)


_data/nav.yml
```diff
@@ -17,7 +17,7 @@ mobile:
   
 
 UI components:
-  - text: Overview
+  - text: UI components
```
shawnbot: You can actually nix the `text:` property altogether. This ensures that if we change the page title in its frontmatter we won't have to fix it here, too.

juliaelman: Doh! Fixed here bb186fb

PR #153, opened 2017-02-22
Change What's new nav structure to use curly quotes

juliaelman:


juliaelman:
@shawnbot updated here 69f90df

juliaelman:
@shawnbot took a closer look and remembered my rationale for the original approach. The left nav odes not compile when using the HTML entity vs. code. I haven't taken a look at the plugin for the nav, but am thinking we leave as character codes for now. Open to suggestions!

PR #165, opened 2017-02-23
Add home calls to action

### LINKED ISSUE #147: Add calls to action to the service offerings
The [training and service offerings](https://standards-staging.usa.gov/#new-training-and-services) on our home page are pretty lonely down there below the fold and don't include any links for more information. I suggest that we add either a single CTA link (a la in the footer) for general service inquiries, or one each for training and customization with pre-filled subject lines.

shawnbot:
Fixes #147. I've added button-like links to the home page for our services:

image

Along the way I also tweaked the icon-in-a-button styles in both these email CTAs and the footer links. Before (left) vs. after (right):

image

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -43,20 +43,20 @@
       </div>
       <div class="usa-footer-contact-links usa-width-one-third">
         <h5>Find an issue or have a feature request?</h5>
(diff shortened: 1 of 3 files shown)


bradnunnally:
LGTM! Great job ya'll!

Issue #185, opened 2017-02-27
sidenav safari bug


bruffridge:
When the browser height is not big enough to contain the sidenav it runs up into the header. See https://standards.usa.gov/components/

<img width="856" alt="screen shot 2017-02-27 at 2 28 10 pm" src="https://cloud.githubusercontent.com/assets/1322063/23376161/2d18fe8c-fcf9-11e6-9a3a-a4617e6e49cc.png">

juliaelman:
@bruffridge thank you for opening this issue! We're working on a solve right now and appreciate you pointing this out to us for review.

shawnbot:
Fixed in 18F/web-design-standards-docs#190. We'll be deploying this to production later today.

PR #195, opened 2017-02-28
Deploy various fixes to production

shawnbot:
This PR will deploy the following changes:

* Add trailing period to heading on the principles page ( #189 )
* Address layout issues with the sticky sidenav ( #185 )
* Add a redirect for the old product roadmap URL ( #191 )
* Update location of the HTML components ( #193 )

Kudos to @mjankowski for helping us out with these!


juliaelman:
@shawnbot 🎉 

Holding off from merging this in until we can test on staging - which is currently not deploying b/c of today's AWS s3 💥

Issue #201, opened 2017-03-02
Disable form submission on component examples


juliaelman:
An example of this is on the https://standards.usa.gov/components/form-templates/#sign-in-form. This expected behavior is that the form would not submit, nor cause any changes in the url structure. 

This is also part of the work to complete the move over to Federalist.

shawnbot:
Fixed in #202.

PR #203, opened 2017-03-04
Deploy to Federalist, URL redirects, and other fixes

shawnbot:
As part of this "release" of the site (not the Standards), we'll also need to do the following:

- [x] Set the custom domain in [Federalist](https://federalist.fr.cloud.gov) to `https://standards.usa.gov`
- [x] Wait for the site to rebuild
- [x] Switch over the DNS

In this release of the site:

* The site will build on Federalist from here on out—specifically, the so-called "[new Federalist](https://federalist.fr.cloud.gov)" on AWS GovCloud
* Future commits to our `master` and `staging` branches will not automatically deploy to cloud.gov
* Disables example form submissions; see: #201 
* Fixes links back to the page template docs from demo pages (thanks, @bruffridge!)


juliaelman:
Noting here, I set the custom domain to `https://standards.usa.gov` and it looks like the build works  🎉

juliaelman:
@wslack not sure if this is the right place to do this (Slack is down ATM), but we are 👍  on setting up the custom domain for the Standards to host on Federalist 🎉  

Just read the docs and it looks like this is the next step:

> The Federalist team uses the cloud.gov CloudFront broker to begin set up for a distribution for a given URL.

wslack:
@juliaelman y'all ready to make the DNS change? Should the DNS controls move to https://github.com/18F/dns?

I can start the process on my side but we need to know the DNS change process before starting.

juliaelman:
@wslack good questions! I've reached out to Russell O'Neil about our DNS switch, as he can help make that change. Our DNS settings are a bit different than other projects and will let you know ASAP what the next steps on that will be.

wslack:
`cf create-service cdn-route cdn-route standards.usa.gov-route -c '{"domain": "standards.usa.gov", "origin": "federalist-proxy.app.cloud.gov", "path": "/site/18f/web-design-standards-docs"}'`

Ready to run this command when y'all are ready to set DNS!

juliaelman:
**UPDATE:** Talked with @wslack and we've run this command 🚀  Handed over the url to Russell to update the CNAME record. We should be moved over to Federalist within the next 24 hours and will update that once confirmed.

wslack:
https://standards.usa.gov/ is LIVE!

wslack:
Wait.........how did you deploy this without merging? Is Federalist using `staging` as the default branch?

PR #216, opened 2017-03-21
Created a retired design file section

bradnunnally:
Moved the omnigraffle download link to a new section on the download
page and informed users that they have been retired.


bradnunnally:
LGTM!

PR #223, opened 2017-03-28
Update: Analytics Q1 2017

shawnbot:
~Currently, this publishes the analytics numbers from October - December, 2016. I think we should consider publishing more recent numbers once March is over.~

Brad has updated this with the (calendar year) Q1 2017 stats, and added epa.gov to our list of domains collected.


bradnunnally:
@shawnbot I've updated the numbers as we discussed yesterday. One thing that I caught was that EPA was missing from the regex code. I added it, but I also added a task for us to go through the regex code and DAP to make sure all known WDS sites are included in the code. That issue is https://github.com/18F/web-design-standards/issues/1821

PR #230, opened 2017-03-29
Fix skip nav link focus issues in IE11

shawnbot:
[👀 Preview on Federalist](https://federalist.fr.cloud.gov/preview/18f/web-design-standards-docs/patch-main-tabindex/)

This is a fix for 18F/web-design-standards#1788 that adds `tabindex="-1"` to `<main>` elements throughout the site, which ensures that IE11 (and earlier?) properly focuses the target of the skip navigation link.

### DIFF
--- a/_components/page-templates/template-docs.html
+++ b/_components/page-templates/template-docs.html
@@ -116,7 +116,7 @@ layout: demo-templates
       </div>
     </header>
     <div class="usa-overlay"></div>
-    <main class="usa-grid usa-section usa-content usa-layout-docs" id="main-content">
+    <main class="usa-grid usa-section usa-content usa-layout-docs" id="main-content" tabindex="-1">
       <aside class="usa-width-one-fourth usa-layout-docs-sidenav">
         <ul class="usa-sidenav-list">
           <li>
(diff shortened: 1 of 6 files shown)


donjo:
One thing this seems to do is it also makes the content areas clickable and when you click anywhere in the div, it puts a border around the content. 

Is there a way to avoid this (for cosmetic reasons) while still being accessible in IE? 

screen shot 2017-03-30 at 12 12 05 pm

msecret:
I usually read that you shouldn't disable outline styles on focus for good accessibility. But I guess in this case, if the `tabIndex` is `-1` then it shouldn't be focused at all?

nickbristow:
Normally you shouldn't disable outline on focused things. The only thing I'd disable it on is, target elements. You don't actually interact with it and when you click an anchor tag a user isn't looking to be focused on an interactive element. Anything in the normal tab index needs focus outlines.

nickbristow:
@shawnbot yeah i think doing blanket CSS like that is probably not the right move for every situation we could predict.

donjo:
@shawnbot well it seems like it's visible to everyone if they happen to click on any of the text that's wrapped in the `<div>` with a `tabindex="-1"` which isn't something most people probably do but I noticed it when I was clicking on my browser window to refocus and I double-clicked. 

To me it creates a false sense of the content being something that can be interacted with so if styling them with a `outline: none` is a no go, I'd probably lean in favor of leaving them out.

PR #250, opened 2017-04-11
[WIP] Add performance pages

shawnbot:
This pull request adds placeholder pages for our web performance work (links should take you to Federalist previews):

1. [Overview](https://federalist.fr.cloud.gov/preview/18f/web-design-standards-docs/add-performance-section/performance/)
1. [Research](https://federalist.fr.cloud.gov/preview/18f/web-design-standards-docs/add-performance-section/performance/research/)
1. [Metrics](https://federalist.fr.cloud.gov/preview/18f/web-design-standards-docs/add-performance-section/performance/metrics/)
1. [Tools](https://federalist.fr.cloud.gov/preview/18f/web-design-standards-docs/add-performance-section/performance/tools/)

A couple of things I'd love feedback on:

1. Is breaking up Metrics and Tools overkill?
1. For consistency, should the URL paths read `/getting-started/performance/`, or is it okay if they're just `/performance/`?
1. The nav logic for highlighting the current page and showing the subnav isn't working for the Performance section, which is the first page on our site with actual "child" pages, as opposed to just in-page anchor links. ☹️  Should I hack in a solution to keep the ball rolling, or take the time to Do It Right?


msecret:
So this was my idea of how to break out:

- What is performance
- Why track performance
- How to track performance
- Performance glossary
- HTTP/2 performance guide
- Research report

msecret:
So we currently don't have anything written for tools, so I'm a little concerned of having it as a tab until we do.

msecret:
So one thing to discuss: the performance writings are pretty large documents right now, and could probably benefit from their own navigation to jump around to things. For instance, it'd help a lot if the web metrics glossary (or whatever we call it) has navigation so a user could skip to different types of metrics.

PR #255, opened 2017-04-14
New Homepage

donjo:
### [:eyes: Preview on Federalist](https://federalist.fr.cloud.gov/preview/18f/web-design-standards-docs/new-homepage/)

The post-1.0 homepage to go out with 1.1.0


css/styleguide.scss
```diff
@@ -10,14 +10,95 @@ body {
...
+  margin-bottom: 1rem;
+}
+
+.usa-section-dark p {
+  margin-top: 0.8rem;
+  font-weight: $font-light;
+}
+
+// Override for 4x1 media grid when it collapses to 2x2
+@media screen and (max-width: 1200px) and (min-width: 600px) {
```
toolness: Interesting, there aren't any SASS variables we can reuse instead of hard-coding `1200px` and `600px` here?  Just curious, I'm not very familiar w/ the USWDS SASS/CSS yet.

toolness:
Um, I don't seem to have the ability to *re*-review a PR once I've dismissed it, so I guess I'll just leave a comment here...

I looked at the code and it looks good!  I say 🚢 if you are OK with the smooshed diamond on IE11.  I've filed the footer bug separately as #331.

PR #279, opened 2017-04-28
Fix code (and other) accordions

### LINKED ISSUE #278: Accordion toggle behavior broken on form templates page
from Slack:

> The accordion toggle behavior appears to be broken on the doc pages at standards.usa.gov (https://standards.usa.gov/components/form-templates/ is what I'm currently on). Double-checked that I have JS enabled. Safari 10.1 + OS X 10.12.4

shawnbot:
### [:eyes: Preview on Federalist](https://federalist.fr.cloud.gov/preview/18f/web-design-standards-docs/patch-accordion-ids/)

This fixes #278 and https://github.com/18F/web-design-standards/issues/1225. In this PR:

* Fix code accordions generally by [replacing](https://github.com/18F/web-design-standards-docs/commit/027ae0636507c473bf0676e71ecfc6177365bea5) the `usa-button-unstyled` class with `usa-accordion-button`.
* Replace all instances of `collapsible-0` and similar IDs with more descriptive ones for each component example.
* Add missing submit buttons to some form examples that lacked them.
* Re-indent code examples to column zero, strip leading and trailing newlines, and remove trailing slashes in self-closing HTML elements (`<input>` not `<input/>`, etc.).
* Exclude `/components/all/` page from output, since we're not referencing it anywhere right now.
* Upgrade [jekyll-redirect-from](https://github.com/jekyll/jekyll-redirect-from) to get the better [redirect template](https://github.com/jekyll/jekyll-redirect-from/blob/master/lib/jekyll-redirect-from/redirect.html).

To test:

1. Visit [the form controls page](https://federalist.fr.cloud.gov/preview/18f/web-design-standards-docs/patch-accordion-ids/components/form-controls/).
1. Click on any "Code" accordion to open and close it.
1. **They should all work now!** If they don't, please let me know.

### DIFF
--- a/css/styleguide.scss
+++ b/css/styleguide.scss
@@ -946,7 +946,7 @@ code {
   }
 
   .usa-accordion-content {
-    padding: 0;
+    // padding: 0;
   }
 }
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,12 +1,15 @@
 GEM
   remote: https://rubygems.org/
   specs:
+    addressable (2.5.1)
+      public_suffix (~> 2.0, >= 2.0.2)
     colorator (1.1.0)
-    ffi (1.9.14)
+    ffi (1.9.18)
     forwardable-extended (2.6.0)
     hash-joiner (0.0.7)
       safe_yaml
-    jekyll (3.2.1)
+    jekyll (3.4.3)
+      addressable (~> 2.4)
       colorator (~> 1.0)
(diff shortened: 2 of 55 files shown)


css/styleguide.scss
```diff
@@ -946,7 +946,7 @@ code {
   }
 
   .usa-accordion-content {
-    padding: 0;
+    // padding: 0;
```
msecret: Any reason to not just remove?

PR #298, opened 2017-05-17
adding paragraph width class to example alerts

donjo:
This adds a class to the alerts used in the performance guide to ensure that they aren't wider than the paragraph content.

The class is on track to be added to the next version of the standards via 18F/web-design-standards#1925

### DIFF
--- a/_includes/perf_example.html
+++ b/_includes/perf_example.html
@@ -1,8 +1,8 @@
-<div class="usa-alert usa-alert-info">
+<div class="usa-alert usa-alert-info usa-alert-paragraph">
   <div class="usa-alert-body">
-    <h3 class="usa-alert-heading">example</h3>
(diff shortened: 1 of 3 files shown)


_includes/perf_example.html
```diff
@@ -1,8 +1,8 @@
-<div class="usa-alert usa-alert-info">
+<div class="usa-alert usa-alert-info usa-alert-paragraph">
   <div class="usa-alert-body">
-    <h3 class="usa-alert-heading">example</h3>
+    <h3 class="usa-alert-heading">Example</h3>
```
msecret: Are we going to keep this heading for now?

donjo: Yeah, for now at least. We should circle back and make a better version of the alerts at a later date.

PR #318, opened 2017-06-19
USAJOBS case study

donjo:
Adding the USAJOBS case study


AndreaSigz:
@donjo looks good to me. Thx!

package.json
```diff
@@ -81,7 +81,7 @@
     "nswatch": "^0.2.0",
     "politespace": "^0.1.4",
     "run-sequence": "^1.1.5",
-    "uswds": "1.2.0-rc1",
+    "uswds": "1.2.1",
```
toolness: Is this change related to the USAJOBS case study, or was it just convenient to upgrade `uswds` in this PR?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "uswds-docs",
-  "version": "1.2.0",
+  "version": "1.2.1",
```
toolness: Are we upgrading the version here because we've upgraded the `uswds` dependency requirement, or is it because the USAJOBS case study constitutes a new version of the docs?

donjo: actually yeah that's a good question. Up until now I think the site has just been matching versions with the standards since we'd tend to push updates to both at the same time but I'm thinking we should probably figure out a better pattern for versioning the site going forward (like does a content update constitute a version bump?)

package.json
donjo: The latter, but going forward I should break this out into an individual PR.

package.json
toolness: Oh, good question. For CALC we only versioned on pushes from `develop` to `master`... So it's similar to your current workflow, I think, only that we do the version bump in PRs like #320, and we also update the [`CHANGELOG.md`](https://calc.gsa.gov/docs/changelog.html) in that PR too.  It's documented in our [release process](https://calc.gsa.gov/docs/release.html) which is a little more heavyweight than I'd like it to be, I guess, but it works.

PR #321, opened 2017-06-26
Add basic site crawler to find 404s, etc.

### LINKED ISSUE #269: Add site tests
Before we can proceed with #130, we need to put some tests in place to:

- [ ] Generate a list of the existing URLs
- [ ] Ensure that we don't break any links in that list
- [ ] Ensure that our HTML is valid
- [ ] Ensure that no images or links are broken
- [ ] Run other basic a11y validations on all pages?

toolness:
This is a **work-in-progress** attempt to fix #269, or at least part of it.

## Instructions

1. Generate the site using `jekyll build`.  The `_site` folder should now contain the latest version of the site.

2. Run `npm run crawl`.  It will let you know if it found any errors.

## Notes

I decided to use [`node-simplecrawler`](https://github.com/cgiffard/node-simplecrawler) because I've had experience using it in the past, and it seems reasonably fast and extensible.

Unlike https://github.com/18F/content-guide/pull/132, I decided not to go with [`html-proofer`](https://github.com/gjtorikian/html-proofer) because it's Ruby-based, and I don't have a ton of experience with Ruby. I also heard rumors that we might consider migrating this site from Jekyll to Hugo; if that happens, we would likely remove Ruby entirely from this project, so I figured node might be a safer long-term option. We can always switch, though!

## To do

Some of these can be filed as separate issues and dealt with in separate PRs.

- [x] Consider making the crawler crawl more than just the index page and all the immediate resources it links to.
- [x] Fix 404s reported by the tool, or log warnings if the 404s come from external sources. Currently these seem to be:
  * ~~All the JS scripts referenced by `ie-polyfill-scripts.html`.~~
  * ~~`html5shiv.js`, which is in an IE-only comment in `head.html`.~~
  * ~~In the developer's guide, there's a broken link to `CONTRIBUTING.md`.~~
- [x] Run the crawler as part of `npm test` (and therefore during CI).
- [x] Don't report errors until we're done w/ the crawl; this way our list of referrers is accurate.

### DIFF
--- a/config/crawl.js
+++ b/config/crawl.js
@@ -1,10 +1,10 @@
+const path = require('path');
+const fs = require('fs');
 const express = require('express');
 const Crawler = require("simplecrawler");
 const chalk = require('chalk');
 
 const app = express();
-
-app.use(express.static(`${__dirname}/../_site`));
 
 // These pages incorporate content from other files in other repos, so
 // they should be considered "second class" by the link checker, and
@@ -16,6 +16,7 @@ const WARNING_PAGES = [
 ];
 const WARNING = chalk.yellow('Warning');
 const ERROR = chalk.red('Error');
+const SITE_PATH = path.normalize(`${__dirname}/../_site`);
 
 function shouldFetch(item, referrerItem) {
   if (item.path.match(/&quot;/)) {
@@ -30,6 +31,13 @@ function shouldFetch(item, referrerItem) {
   }
 
   return true;
+}
+
+if (fs.existsSync(SITE_PATH)) {
+  app.use(express.static(SITE_PATH));
+} else {
+  console.log(`Please build the site before running me.`);
+  process.exit(1);
 }
 
 const listener = app.listen(() => {
--- a/config/crawl.js
+++ b/config/crawl.js
@@ -3,7 +3,7 @@ const Crawler = require("simplecrawler");
 
 const app = express();
 
-app.use(express.static(`${__dirname}/_site`));
+app.use(express.static(`${__dirname}/../_site`));
 
 // These pages incorporate content from other files in other repos, so
 // they should be considered "second class" by the link checker, and
--- a/config/crawl.js
+++ b/config/crawl.js
@@ -1,5 +1,6 @@
 const express = require('express');
 const Crawler = require("simplecrawler");
+const chalk = require('chalk');
 
 const app = express();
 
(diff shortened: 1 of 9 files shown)


toolness:
@donjo any idea what to do about these "less than IE9"-specific JS files that don't seem to exist?  Do we even support < IE9 anymore?  If not, I guess we can just get rid of the references...

toolness:
Hmm, @donjo just mentioned on slack that our [documentation page on accessibility](https://standards.usa.gov/getting-started/developers/#notes-on-accessibility) mentions that we only support IE9 and above.

toolness:
So I'm finding that a number of 404s are actually coming from external sources, like `WHO_IS_USING_USWDS.md` and the release notes, which are stored in the other repository. For these I'm thinking maybe we should log *warnings* but not errors, so that a broken link caused by one of those files doesn't break the whole build.

donjo:
➕  to warnings for external source 404s. Agree.

config/crawl.js
```diff
@@ -0,0 +1,100 @@
...
+    listener.close(() => {
+      let errors = 0;
+      let warnings = 0;
+
+      notFound.forEach(item => {
+        const refs = referrers[item.url];
+        const isWarning = refs.every(path => WARNING_PAGES.includes(path));
+        const label = isWarning ? WARNING : ERROR;
+
+        console.log(`${label}: 404 for ${item.path}!`);
```
jseppi: leaving off the `!` would clean up the output a bit, and make it easier to copy the path

config/crawl.js
```diff
@@ -0,0 +1,100 @@
+const express = require('express');
+const Crawler = require("simplecrawler");
+const chalk = require('chalk');
+
+const app = express();
+
+app.use(express.static(`${__dirname}/../_site`));
```
jseppi: Maybe check that `_site` has contents, otherwise the call to `refs.every(...)` later on (and potentially other things) will fail.


Here's output from running `yarn crawl` without first running `yarn build`:

```sh
/Users/jamesseppi/CODE/web-design-standards-docs/config/crawl.js:67
        const isWarning = refs.every(path => WARNING_PAGES.includes(path));
                              ^

TypeError: Cannot read property 'every' of undefined
    at notFound.forEach.item (/Users/jamesseppi/CODE/web-design-standards-docs/config/crawl.js:67:31)
```

toolness:
Good suggestions @jseppi! Just incorporated them.

@donjo I think this is good to merge if you are OK with the functionality!

toolness:
woooooooot!

PR #344, opened 2017-07-11
Run aXe on a few pages during CI

### LINKED ISSUE #326: Run a11y checks on site during 'npm test'
This is spun off from #269; from that issue, I wrote:

> Doing a11y checks is also a bit challenging because doing it in a reasonable way requires a headless browser; I've got some code in our [a11y-metrics](https://github.com/18F/a11y-metrics) repo that could help with this, but the checking also takes some time due to the headless browser intermediary, so it might only make sense to run it on a short whitelist of pages (e.g. this "kitchen sink" page I've heard referenced in other issues, not sure where it is though).

toolness:
Fixes #326.

This runs [axe-core](https://github.com/dequelabs/axe-core) on a few pages of the standards during CI and `npm test` runs. In future PRs we can add support for running it on more pages.

A successful run looks something like this:

> 

It requires that Chrome/Chromium v59 or higher be installed on the host system. I decided on [headless Chrome](https://developers.google.com/web/updates/2017/04/headless-chrome) because:

* It's maintained by the Google Chrome team, so it's likely to be very well-maintained going forward. Possibly even moreso than Phantom; I vaguely recall seeing a Tweet from Ariya Hidayat a while back saying that Phantom was going to be end-of-life'd since Chromium is headless is officially supported now, but I could be totally wrong on that.
* I haven't done any formal measurements, but headless Chrome seems _really_ fast compared to Phantom.
* In the vast majority of cases, it's just one less thing for developers to install--every developer I know has Chrome installed on their system, and our use of `chrome-launcher` ensures that it will be found, regardless of host OS.
* I noticed we have a (kind of old) WIP PR to add support for Google Lighthouse in #210, so this PR might make continuing work on that easier too,
* Chrome is a real browser, unlike PhantomJS, so it's likely that any errors can be reproduced and debugged using Chrome running in non-headless (headed?) mode.

To do:
- [x] Ensure this works on CircleCI.
- [x] Ensure we're running aXe on the actual front page and not e.g. a "network error" page.
- [x] Ensure that [`docker-uswds`](https://github.com/toolness/docker-uswds) can run this too.
- [x] Update README to indicate that developers should have Chrome/Chromium installed on their systems for tests to run properly.
- [x] Modify `crawl.js` to use `static-server.js` to prevent duplication of code.

### DIFF
--- a/config/run-axe.js
+++ b/config/run-axe.js
@@ -41,14 +41,14 @@ function getRemoteChrome() {
 // This function is only here so it can be easily .toString()'d
 // and run in the context of a web page by Chrome. It will not
 // be run in the node context.
-function runAxe() {
+const RUN_AXE_FUNC_JS = function runAxe() {
   return new Promise((resolve, reject) => {
     window.axe.run((err, results) => {
       if (err) return reject(err);
       resolve(JSON.stringify(results.violations));
     });
   });
-}
+}.toString();
 
 let getChrome = REMOTE_CHROME_URL ? getRemoteChrome : launchChromeLocally;
 
@@ -59,7 +59,7 @@ Promise.all([runServer(), getChrome()]).then(([server, chrome]) => {
   console.log(`Running aXe on:`);
 
   CDP({
-    host: chrome.host,
+    host: chromeHost,
     port: chrome.port,
   }, client => {
     const {Page, Network, Runtime} = client;
@@ -110,7 +110,7 @@ Promise.all([runServer(), getChrome()]).then(([server, chrome]) => {
       });
       Page.loadEventFired(() => {
         Runtime.evaluate({
-          expression: AXE_JS + ';(' + runAxe.toString() + ')()',
+          expression: `${AXE_JS};(${RUN_AXE_FUNC_JS})()`,
           awaitPromise: true,
         }).then(details => {
           let errorFound = false;
@@ -128,7 +128,7 @@ Promise.all([runServer(), getChrome()]).then(([server, chrome]) => {
               errorFound = true;
             }
           } else {
-            console.log('Unexpected result from CDP!');
+            console.log('Unexpected result from aXe JS evaluation!');
             console.log(details.result);
             errorFound = true;
           }
@@ -149,4 +149,8 @@ Promise.all([runServer(), getChrome()]).then(([server, chrome]) => {
       loadNextPage();
     });
   });
+}).catch(e => {
+  console.log(chalk.red("An error occurred during initialization."));
+  console.log(e);
+  process.exit(1);
 });
--- a/config/run-axe.js
+++ b/config/run-axe.js
@@ -4,6 +4,7 @@ const fs = require('fs');
 const urlParse = require('url').parse;
 const chromeLauncher = require('chrome-launcher');
 const CDP = require('chrome-remote-interface');
+const chalk = require('chalk');
 const runServer = require('./static-server');
 
 const REMOTE_CHROME_URL = process.env['REMOTE_CHROME_URL'];
@@ -55,6 +56,7 @@ Promise.all([runServer(), getChrome()]).then(([server, chrome]) => {
   const chromeHost = chrome.host || 'localhost';
   console.log(`Static file server is listening at ${server.url}.`);
   console.log(`Chrome is debuggable on http://${chromeHost}:${chrome.port}.`);
+  console.log(`Running aXe on:`);
 
   CDP({
     host: chrome.host,
@@ -69,7 +71,7 @@ Promise.all([runServer(), getChrome()]).then(([server, chrome]) => {
       } else {
         const page = pagesLeft.pop();
         const url = `${server.url}${page}`;
-        console.log(`Navigating to ${page}.`);
+        process.stdout.write(`  ${page} `);
         Page.navigate({url});
       }
     };
@@ -80,7 +82,8 @@ Promise.all([runServer(), getChrome()]).then(([server, chrome]) => {
           // the remote chrome instance (if we're using one) may be
           // keeping some network connections to the server alive, which
           // makes it harder to kill, so it's easier to just terminate.
-          console.log(`Terminating with exit code ${exitCode}.`);
+          const color = exitCode === 0 ? chalk.green : chalk.red;
+          console.log(color(`Terminating with exit code ${exitCode}.`));
           process.exit(exitCode);
         });
       });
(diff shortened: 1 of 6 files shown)


toolness:
Hmm, so this is a fair bit of code, as the [Chrome DevTools Protocol](https://chromedevtools.github.io/devtools-protocol/) is fairly low-level. There might be a higher-level wrapper available that could reduce the size of our code, though I'm not sure if it'd necessarily meet our particular needs or not. Even axe-core's own PhantomJS wrapper didn't meet all the needs for the [a11y-metrics](https://github.com/18F/a11y-metrics) dashboard so I had to write my own.

config/static-server.js
```diff
@@ -0,0 +1,30 @@
...
+
+if (fs.existsSync(SITE_PATH)) {
+  app.use(express.static(SITE_PATH));
+} else {
+  console.log(`Please build the site before running me.`);
+  process.exit(1);
+}
+
+module.exports = () => {
+  return new Promise((resolve, reject) => {
```
msecret: Does this need a rejection handler somewhere?

config/run-axe.js
```diff
@@ -0,0 +1,152 @@
...
+  });
+}
+
+let getChrome = REMOTE_CHROME_URL ? getRemoteChrome : launchChromeLocally;
+
+Promise.all([runServer(), getChrome()]).then(([server, chrome]) => {
+  const chromeHost = chrome.host || 'localhost';
+  console.log(`Static file server is listening at ${server.url}.`);
+  console.log(`Chrome is debuggable on http://${chromeHost}:${chrome.port}.`);
+  console.log(`Running aXe on:`);
```
msecret: Was there more text that was supposed to go here?

config/run-axe.js
```diff
@@ -0,0 +1,152 @@
...
+
+  CDP({
+    host: chrome.host,
```
msecret: Above, it sounds like `chrome.host` may not exist. Should this be `chromeHost`? Why wouldn't `chrome.host` not exist anyway?

config/run-axe.js
```diff
@@ -0,0 +1,152 @@
...
+    const {Page, Network, Runtime} = client;
+    const pagesLeft = PAGES.slice();
+    const loadNextPage = () => {
+      if (pagesLeft.length === 0) {
+        console.log(`Finished visiting ${PAGES.length} pages with no errors.`);
+        terminate(0);
+      } else {
+        const page = pagesLeft.pop();
+        const url = `${server.url}${page}`;
+        process.stdout.write(`  ${page} `);
```
msecret: What's this doing? Why isn't it a simple `console.log`?

config/run-axe.js
```diff
@@ -0,0 +1,152 @@
...
+    process.on('unhandledRejection', (reason, p) => {
+      console.log('Unhandled Rejection at:', p, 'reason:', reason);
+      terminate(1);
+    });
+
+    Promise.all([
+      Page.enable(),
+      Network.enable(),
+    ]).then(() => {
+      Network.responseReceived(({response}) => {
```
msecret: Is this when any page resource is received? Or just the main HTML document?

config/run-axe.js
```diff
@@ -0,0 +1,152 @@
...
+        terminate(1);
+      });
+      Network.loadingFailed(details => {
+        console.log("A network request failed to load.");
+        console.log(details);
+        terminate(1);
+      });
+      Page.loadEventFired(() => {
+        Runtime.evaluate({
+          expression: AXE_JS + ';(' + runAxe.toString() + ')()',
```
msecret: Is this possible to write using a template string so it's easier to parse?

config/run-axe.js
```diff
@@ -0,0 +1,152 @@
...
+              console.log(chalk.red(`Found ${viols.length} aXe violations.`));
+              console.log(viols);
+              console.log(chalk.cyan(
+                `\nTo debug these violations, install aXe at:\n\n` +
+                `  https://www.deque.com/products/axe/\n`
+              ));
+              errorFound = true;
+            }
+          } else {
+            console.log('Unexpected result from CDP!');
```
msecret: What's CDP?

config/static-server.js
toolness: Ah good CATCH, will add!

config/run-axe.js
toolness: Nah, it's just the heading text that appears before each individual page that's visited. The final output looks like this:

config/run-axe.js
toolness: Oh, I wanted it to look cool like all the fancy test runners!

config/run-axe.js
toolness: Oh, it stands for Chrome DevTools Protocol... I'll expand it to just say that!

config/run-axe.js
toolness: I believe it's when *any* page resource is received.  So this effectively overlaps a bit with our crawler from #321.  It was easier to do it that way than it was to check the URL and see if it matched the page we were requesting, but I could add such a conditional if you think it'd be better that way.

config/run-axe.js
toolness: Oh, it will just be `undefined` if it doesn't exit, which CDP just then defaults to localhost--but I agree, I think it'd be less confusing / more explicit if we just passed `chromeHost` here.

I was hoping that the API for `chrome` would set `host` to whatever the ultimate "resolved" value was, whether that was `localhost` (in the case where Chrome is launched locally) or something explicit (in the case where we connect to a remote Chrome in e.g. a different Docker container).  However, that wasn't the case--in the case where Chrome is launched locally, `host` isn't set at all, so I had to manually default it to `localhost` via that `chromeHost` variable. It's a bit annoying because the APIs that connect all these different Chrome packages isn't quite as consistent as I'd hoped, and this is a workaround for that.

config/run-axe.js
msecret: Seems OK to have this extra check

toolness: Hmm, so I thought about it a bit, and realized that maybe it's actually good that it's counting any sub-resource failure as a 404 too.  For example, if it's 404'ing on CSS or JS, then that changes the accessibility profile of the page--calculated color contrast ratios may be incorrect, or certain ARIA attributes may not be present on the evaluated page.

Another thing is that the crawler from #321 isn't actually 100% accurate--the library we're using uses regular expressions to "scrape" resources for URLs, so it's actually possible that it might miss sub-resources that are noticed by actually loading a page in Chrome. So it might still be good to hard fail on sub-resource failures for the sake of redundancy, too.

What do you think? At the very least, I think we should log warnings on sub-resource fails.

toolness:
Ok, I think I addressed all your issues, though I had second thoughts about ignoring sub-resource loading failures, which I [commented on above](https://github.com/18F/web-design-standards-docs/pull/344#discussion_r127084792).

config/run-axe.js
msecret: Yea I agree that it's good to consider sub-resource 404s as failures.

toolness:
Hooray, thanks! 💃

PR #348, opened 2017-07-11
JS customization documentation

msecret:
## Description

Adds documentation to developer guide on how to extend and use the JavaScript. 

I removed outdated information about jQuery as well, I mistakenly committed it to the develop branch here: https://github.com/18F/web-design-standards-docs/commit/bafe664640ae2bb024c576f3b666bff4b4ed11de

Fixes https://github.com/18F/web-design-standards/issues/1975

## Additional information

- I found through investigation that our components do not yet emit events when actions happen, which would greatly improve the API to extend them. This is something that should be looked into.
- The USWDS JS code is not exposed in a global scope at all, meaning it's very hard to extend from it if developers are not using an npm/build CommonJS module workflow. This is another thing to look into and would be an easy fix.
- Our JS code has very little documentation, which could also be improved.

### DIFF
--- a/pages/getting-started/developers.md
+++ b/pages/getting-started/developers.md
@@ -148,6 +148,16 @@ The JavaScript for the standards is separated into components in the same manner
 window.uswds = require('./components');
 ```
 
+Each component has a standardized interface that can be used to extend it further. The primary methods each component has are as follows:
+
+- `on`: Initialize a component's JavaScript behavior by passing the root element, such as `window.document`.
+- `off`: The opposite of `on`, de-initializes a component, removing any JavaScript event handlers on the component.
+- `hide`: Hide the whole component.
+- `show`: Shows a whole, hidden component.
+- `toggle`: Toggles the visibility of a component on and off based on the previous state.
+
+Some components have additional methods for manipulating specific aspects of them based on what they are and what they do. These can be found in the component's JS file.
+
 ## Customization and theming
 
 The Standards can be customized to use different typography, colors and grid systems. The easiest way to do this is to use Sass and override the Standards’ global variables. If it isn’t possible to use Sass, do theming by overriding the CSS rules in the Standards set.
--- a/pages/getting-started/developers.md
+++ b/pages/getting-started/developers.md
@@ -150,7 +150,7 @@ The JavaScript for the standards is separated into components in the same manner
 window.uswds = require('./components');
 ```
 
-Each component has a standardized interface that can be used to extend it further. The primary methods each component has are as follows:
+Each component has a standardized interface that can be used to extend it further. The components store a HTML classname that's used to link HTML elements with the JS component, so when a component is initialized, it will search through the current HTML DOM finding all elements that match it's class and inialize the component JavaScript for those elements. The primary methods each component has are as follows:
 
 - `on`: Initialize a component's JavaScript behavior by passing the root element, such as `window.document`.
 - `off`: The opposite of `on`, de-initializes a component, removing any JavaScript event handlers on the component.
--- a/pages/getting-started/developers.md
+++ b/pages/getting-started/developers.md
@@ -142,7 +142,7 @@ You can also email us directly at [uswebdesignstandards@gsa.gov](mailto:uswebdes
 
 ## JS customization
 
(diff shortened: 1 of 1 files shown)


pages/getting-started/developers.md
```diff
@@ -140,6 +140,24 @@ You can also email us directly at [uswebdesignstandards@gsa.gov](mailto:uswebdes
 **For more information, visit:
 [https://pages.18f.gov/frontend/css-coding-styleguide/](https://pages.18f.gov/frontend/css-coding-styleguide/)**
 
+## JS customization
+
+The JavaScript for the standards is separated into components in the same manner as the visual interface which is all initialized with event handlers when the DOM is ready. These components are accessible as CommonJS modules that can be required in other JavaScript files which then must be fuilt for the browser. The components are currently not accessible in the global browser scope, but can be easily extended to be included by requiring `components` and setting it to a global scope:
```
toolness: `fuilt` -> `built`, I think?

pages/getting-started/developers.md
```diff
@@ -140,6 +140,24 @@ You can also email us directly at [uswebdesignstandards@gsa.gov](mailto:uswebdes
...
+
+```js
+window.uswds = require('./components');
+```
+
+Each component has a standardized interface that can be used to extend it further. The primary methods each component has are as follows:
+
+- `on`: Initialize a component's JavaScript behavior by passing the root element, such as `window.document`.
```
toolness: It might be worth mentioning something about the fact that we're using a delegation model here, for folks who are unfamiliar with it?  Otherwise readers might get confused, e.g. wondering why anyone would want to turn `window.document` into an accordion...

pages/getting-started/developers.md
toolness: I think we might also want to get rid of the word "easily"...  Aside from our content guild generally discouraging such words, I don't think it's actually that easy given that someone has to set up their own browserify/webpack build system to do this, which can be really frustrating and time-consuming, especially for folks who are unfamiliar with such tools to begin with.

toolness: I wonder if it might be useful to just telegraph the fact that customizing USWDS's JS currently *requires* a tool like browserify or webpack?  E.g. a big bold paragraph at the beginning of this section that says:

> **Unfortunately, customizing the JavaScript for the standards currently requires NodeJS and a module bundler like Browserify or Webpack. We apologize for this inconvenience, and are working to resolve it in a future release of the Standards.**

Or something like that?  I'm concerned that folks who might not be super familiar with JS, e.g. who might not know exactly what "accessible in global browser scope" mean but who do know how to call functions and such, might be confused here. I mean, they might also not really know what node/babel/webpack are, but at least an apologetic tone could help convey the notion that leveraging our JS requires relatively advanced skills right now, especially compared to systems like Bootstrap and Foundation, and we're trying to lower the barrier to entry.

toolness:
This is looking good so far!  It's making me want to migrate the components to just be web components (#1706) because this will actually make documenting them a lot easier, I think.

pages/getting-started/developers.md
```diff
@@ -142,13 +142,15 @@ You can also email us directly at [uswebdesignstandards@gsa.gov](mailto:uswebdes
...
+
+The JavaScript for the standards is separated into components in the same manner as the visual interface which is all initialized with event handlers when the DOM is ready. These components are accessible as CommonJS modules that can be required in other JavaScript files which then must be built for the browser. The components are currently not accessible in the global browser scope, but can be extended to be included by requiring `components` and setting it to a global scope:
 
 ```js
 window.uswds = require('./components');
 ```
 
-Each component has a standardized interface that can be used to extend it further. The primary methods each component has are as follows:
+Each component has a standardized interface that can be used to extend it further. The components store a HTML classname that's used to link HTML elements with the JS component, so when a component is initialized, it will search through the current HTML DOM finding all elements that match it's class and inialize the component JavaScript for those elements. The primary methods each component has are as follows:
```
toolness: Minor typo, `it's -> its`.  Also, should `classname` be `class name`?  It might also be useful to just provide an example in parentheses, like `The components store a HTML class name (e.g. usa-accordion) ...` just to make things a bit more concrete.

msecret:
While it is changing, I'm worried a disclaimer will just cause a loss of trust rather then get people to keep checking back for changes.

toolness:
Ah good point! Yeah, I think as long as we make sure that any breaking changes we make are well-documented in future releases, we should be good.

PR #357, opened 2017-07-19
[WIP] Add {% fractal_component %} liquid template tag

toolness:
This PR is being developed in parallel with https://github.com/18F/web-design-standards/pull/2050.

This helps us with https://github.com/18F/web-design-standards/issues/1765 and makes our code examples more DRY by adding a `{% fractal_component %}` [liquid template tag](https://jekyllrb.com/docs/plugins/#tags) that attempts to pull the rendered HTML of a component into the jekyll site. For example:

```jinja
{% fractal_component accordion--default %}
```

This would find the `accordion--default` component and plop its HTML into the template.

The rendered component is found in one of two ways:

1. If `FRACTAL_BASE_URL` is defined, it's used as the origin of the Fractal dev server, and the component is pulled from it at the `/components/render/:name` endpoint.

2. Otherwise, the `node_modules/uswds` directory is examined for a build of the Fractal static site. Every fractal site (well, at least every Fractal [Mandelbrot](http://fractal.build/guide/web/default-theme)-themed site) contains the raw HTML of rendered components at `components/render/:name.html`, so this file is read.

The first option is preferable for development because it means we can iterate on the Jekyll site *and* the Fractal components it uses without having to constantly rebuild the static Fractal site.

The second option is more useful for production/CI, but it should be noted that **the static version of the Fractal site needs to actually be built in order for it to work**.  We could do this in one of two ways:

1. When we release the npm version of uswds, we also build the static Fractal site and include it in the package. Aside from benefiting the docs, it could also help serve as built-in, offline-ready documentation of the standards for anyone who `npm install`s it, which is nice.

2. We include fractal packages in the docs repository's `package.json` and have the docs build the static fractal site itself if necessary. The benefit of this approach is that it's a bit friendlier to installing uswds from git branches and such, rather than official npm releases--though that approach is already blocked by https://github.com/18F/web-design-standards/issues/2032, so it's not like we'll be breaking anything if we don't follow this approach.

## To do

- [x] Ensure that images referenced by components work (see first comment below for explanation).
- [x] Document the `{% fractal_component %}` tag and associated environment variable(s).
- [ ] Figure out how we want to pull the fractal content during production/CI.


toolness:
It should be noted that there's some limitations/caveats with this approach:

* Right now image assets contained in component snippets, such as the "close button" image in the mobile nav menu, probably won't work, due to relative URL issues.  Fortunately, at least this will be caught by our link-checker that was added in #321. We can probably remedy it by doing a simple find-replace in the plugin--it's not ideal, but it'll get the job done, and we can be confident that any broken links will be found by our checker.

* We need to be extra careful to make sure that some of our Fractal components play well with each other when placed on the same page. For example, while building this plugin, I attempted to replace our accordion example code at [`_includes/code/components/accordion.html`](https://github.com/18F/web-design-standards-docs/blob/develop/_includes/code/components/accordion.html) with the following:

    ```jinja
    <h6>Borderless</h6>
    
    {% fractal_component accordion--default %}
    
    <h6>Bordered</h6>
    
    {% fractal_component accordion--bordered %}
    ```

    However, the second example ended up broken, because our accordions rely on `id` attributes to work properly, and the two components have the same ids, which meant that clicking on the accordions of the second example actually caused the content of the *first* example to be toggled!  I think that some of our components already take this into account, perhaps because they were included on the kitchen sink page or something. So this just means we'll have to apply the same strategy (usually done via an `id_prefix` context key) to more components.

toolness:
Hi @shawnbot!

Hmm, that's an interesting idea... the main drawback I see to it is just the usual iframe height-setting hijinks, but outside of that I see at least one additional benefit: if the rendering is in an iframe, it allows us to have the iframe point directly into the fractal server's component rendering during development, which could make the development workflow better. Right now, changes to the fractal component still require devs to `touch` the jekyll file with the `{% fractal_component %}` tag in it for Jekyll to realize that it needs to rebuild the page, which is annoying. But with an iframe, devs could just make changes to the component and reload the page to see it with their new component. (If we integrate browsersync into it, we could potentially obviate even the need for the reload.)

toolness:
Cool, thanks @shawnbot!

So I've  noticed a few more things after attempting to migrate more components over:

* There are subtle differences between some of our fractal components and our documentation examples. For example, the header examples in the documentation all have the government banner on them, while the ones in the fractal site don't. I'm not sure which route we should go with. Guess we should just decide on a case-by-case basis?

* I think we should try to make sure to preserve the impeccable spacing and indentation of the code in the documentation examples, since they're the items we've got "on display" in our "storefront".  That means e.g. no weird gaps of multiple blank lines, or resetting the indentation level due to the use of a partial. I think there's different ways we can accomplish this but they all involve cleaning things up on the nunjucks side, since that's where the content lives, and that's where it's easiest to iterate on them.

toolness:
Oh neat, the federalist build is now working, and it's using the version of uswds from 18F/web-design-standards#2050, so that's kind of neat.

msecret:
So thinking of iframe: is there any risk that certain government security practices might block iframes from other domains, creating a broken page?

_plugins/fractal_component.rb
```diff
@@ -0,0 +1,52 @@
...
+
+module Jekyll
+  class FractalComponentTag < Liquid::Tag
+    BASE_URL_ENV_VAR = 'FRACTAL_BASE_URL'
+
+    def initialize(tag_name, name, tokens)
+      super
+      @name = name.strip
+      @base_url = ENV[BASE_URL_ENV_VAR]
+      @fs_path = "node_modules/uswds/build/components/render/#{@name}.html"
```
msecret: Is there any danger of this path changing, or components being nested in their own folders here based on how they're nested in the src folders?

_plugins/fractal_component.rb
```diff
@@ -0,0 +1,52 @@
...
+          puts "Loading fractal component #{@name} from #{@fs_path}."
+          html = open(@fs_path).read
+        else
+          raise (
+            "Unable to find the fractal component #{@name}! " +
+            "Please either define #{BASE_URL_ENV_VAR} or run " +
+            "'fractal build' in the uswds directory."
+          )
+        end
+        html.gsub! "../../dist/", "#{site.baseurl}/assets/"
```
msecret: What does this line do?

msecret:
Do we still need a separate math to keep the code samples if they get pulled directly from fractal? So now it goes:

`/components`/ <= `/_includes/code/components/accordion.html` <= fractal

But if all that's in those code component files is the fractal include why not have `/components/` include directly from fractal?

toolness:
Definitely, I think that's the next step!

It might require changing some of the code in the `/_includes/code/` partials, specifically `code/preview.html` and `code/accordion.html`, though, since right now they take a `component` key (such as `component="accordion"`) and use its value to construct a path (such as `/_includes/code/components/accordion.html`), which they then include... It's all kind of dynamic.

config/gulp/build.js
```diff
@@ -38,6 +41,40 @@ gulp.task('clean-assets', function (done) {
...
+
+  if (fs.existsSync(fractalIndex)) {
+    dutil.logMessage('build-uswds-if-needed', 'USWDS is already built.');
+    return Promise.resolve();
+  } else {
+    dutil.logMessage('build-uswds-if-needed', 'Building USWDS...');
+    return new Promise((resolve, reject) => {
+      child_process.spawn(
+          'npm',
+          [ 'run', 'federalist' ],
```
msecret: I'm confused by this. The `web-design-standards` repo doesn't have a `federalist` script, only this docs repo does. Should this be `npm run build` from that directory?

msecret:
I've investigated what it would be like to have the code samples pull directly from fractal as well as the preview:

- Fractal does have a rendering of just the codes samples under `preview`, but it is a full HTML document, so might need to have the `<head>`, `<body>`, etc stripped from it.
- For the button example, I've split up the button into multiple variants. On the docs site preview, I manually layout each variant like so:

```html
<h6>Primary buttons</h6>
<div class="button_wrapper">
  {% fractal_component buttons--primary %}
</div>
<div class="button_wrapper">
  {% fractal_component buttons--primary-alt %}
</div>

<h6>Secondary buttons</h6>
<div class="button_wrapper">
{% fractal_component buttons--secondary %}
</div>
...
```

So to document the buttons, we would either need to include all fractal component includes in the code file:

```html
{% fractal_component_code buttons--primary %}
{% fractal_component_code buttons--primary-alt %}
{% fractal_component_code buttons--secondary %}
...
```

or loop through the variants somehow automatically. I'm not sure which option is better, and could use feedback
- Would it make sense to make a new Liquid tag for the code part of the fractal component? Similar to how I did in the example above.

config/gulp/build.js
toolness: Hmm, I think both do actually--the `web-design-standards`'s `federalist` script generates its static fractal site, which is how we get https://components.standards.usa.gov/ and our federalist branch builds for the standards site.

Should we add a comment that says this, to make it sound less confusing?

toolness:
Oh, for the code samples I was thinking we'd just use the exact same content that we use for `{% fractal_component %}`, we'd just HTML-escape and highlight it on the Jekyll side, similar to what [`_includes/code/syntax.html`](https://github.com/18F/web-design-standards-docs/blob/develop/_includes/code/syntax.html) is doing?  I mean, we want to show the final HTML source for the rendered component rather than the nunjucks source, right?

toolness:
Anyways, yeah, I think a `{% fractal_component_code %}` tag makes sense, only IMO it should just do the HTML-escaping and syntax-highlighting in Jekyll-land rather than trying to pick apart the code samples from the Fractal `preview`...

msecret:
I think I got this kinda working, in a messy way. Please look at last few commits. It's still failing, but at this point, I think it's only failing because it's using a version of the standards that doesn't have the fractal components yet. Should we be pulling the uswds package from master rather then npm?

config/gulp/build.js
```diff
@@ -38,6 +41,51 @@ gulp.task('clean-assets', function (done) {
...
+    dutil.logMessage('build-uswds-if-needed', 'USWDS is already built.');
+    return Promise.resolve();
+  } else {
+    dutil.logMessage('build-uswds-if-needed', 'Building USWDS...');
+
+    //  "federalist": "npm install --dev && gulp build && fractal build",
+    const sharedOpts = { stdio: 'inherit', cwd: uswdsDir };
+    return Promise.all([
+      spawnP('npm', [ 'install', '--only=dev', '--ignore-scripts' ], sharedOpts),
+      spawnP('gulp', ['build', `--cwd ${uswdsDir}` ], sharedOpts),
```
msecret: Run the uswds build commands separately so we can control the `gulp` cwd and stop it from changing to the docs directory

toolness:
Ah yeah I think we should pull npm from master! Imma take a closer look at this now.

_plugins/fractal_component.rb
toolness: Uhh, sorry I never responded to this question... Basically fractal components that have `` tags that reference USWDS's static assets (e.g. the USWDS close mobile nav menu image) will be referencing them at `../../dist`, because that's where the USWDS assets are on the fractal site. However, they're at a different location on the docs site, so we're fixing that up with this line.

It's unintuitive enough (I had to think a while to realize why I'd written it) that I should add a comment explaining it before we merge this!

config/gulp/build.js
toolness: Hmm, I think this is effectively running these three commands in *parallel*, but I'm guessing you want to actually run them in serial, based on the comment in line 79?  I think you might want to change this to something like:

```js
return spawnP('npm', [ 'install', '--only=dev', '--ignore-scripts' ], sharedOpts)
  .then(() => spawnP('gulp', ['build', `--cwd ${uswdsDir}` ], sharedOpts))
  .then(() => spawnP('./node_modules/.bin/fractal', [ 'build' ], sharedOpts));
```

More details on the Promise execution timing at https://stackoverflow.com/a/30823708.

msecret: yea good point, forgot to do that.

msecret:
So another thing to consider here: just we get rid of the `prepublish` hook on uswds?

When trying to look up solutions to this issues yesterday, a very common theme was how confusing the prepublish hook is. I've run into trouble with it before as a developer on WDS. And what is it really accomplishing? The WDS are published in an automated fashion through CI, what is `prepublish` protecting us from?

toolness:
Uh, that's interesting, the `federalist/build` failed but `federalist-staging/build` worked... What's the difference between these two?  I'm unable to view the logs for the former so I can't tell what's wrong.

package.json
```diff
@@ -90,7 +90,7 @@
     "politespace": "^0.1.4",
     "run-sequence": "^1.1.5",
     "simplecrawler": "^1.1.3",
-    "uswds": "1.3.0",
+    "uswds": "18F/web-design-standards#fb49e4f",
```
toolness: Ok, so I'm tempted to say that this is how we should specify the USWDS version--either by a specific commit hash, or by a version tag... in this case I think we need the commit hash because our latest 1.3.0 release might not have all the fractal components we need?

Anyways, i hesitate to put just `18F/web-design-standards` because then we have no idea what version of the standards it's actually using...

toolness:
Heh, because our custom `uswds_version` jekyll plugin is pulling the version string verbatim from `package.json` for inclusion in templates, we're getting pages that look like this:

> version

Guess we could change that so it parses out the part after the `#` and only shows that... hmmmmm. Or we could just make sure to always have `master` use an actual version number for `uswds` in `package.json`.

config/gulp/build.js
```diff
@@ -85,13 +85,8 @@ gulp.task('build-uswds-if-needed', function () {
...
     }
 
-    //  "federalist": "npm install --dev && gulp build && fractal build",
     const sharedOpts = { stdio: 'inherit', cwd: uswdsDir };
-    return spawnP('npm', [ 'install', '--only=dev', '--ignore-scripts' ],
-                  sharedOpts)
-      .then(() => spawnP('gulp', [ 'build' ], sharedOpts))
-      .then(() => spawnP('./node_modules/.bin/fractal', [ 'build' ],
-                         sharedOpts));
+    return spawnP('npm', [ 'run', 'federalist' ], sharedOpts);
```
toolness: @msecret, I'm confused as to why this was changed to the series of 3 commands in the first place?  What problem were you experiencing with just `npm run federalist`?  It seems to work on my end and on `federalist-staging/build`, at least... I think whatever we run, it should delegate as much as possible to an npm script in `uswds`, so that we have as loose a coupling as possible between these two repositories.  Even if that means we add yet another npm script called e.g. `npm run build-fractal-for-uswds-docs` or something nutty.

toolness:
Oof, I tried passing `--verbose` to npm in an effort to debug the problem better, but...

```
ERROR | output suppressed due to length | 2017-08-10T11:22:46.063Z
-- | -- | --
```

toolness:
Ok, I *think* I've fixed it. Basically, when Federalist was running `npm run federalist` in this repo, *that* was running `npm run federalist` inside `node_modules/uswds`, and *that* was running `npm install --dev`, which often resulted in [`Killed`](https://stackoverflow.com/questions/726690/who-killed-my-process-and-why) being logged, and the build failing.

I suspect this was due to `npm install --dev` using up too many system resources but I'm not sure.

I've found that the build on Federalist seems to consistently work if we use node 8, which gives us npm 5, which I think is a lot more stable/efficient/etc than npm 3, which is what comes with node 6.

Before upgrading to node 8 in 93d0684, failed Federalist builds logged this:

```
Now using node v6.11.2 (npm v3.10.10)
```

But now they log this:

```
Now using node v8.3.0 (npm v5.3.0)
```

In particular, looking at the [npm 5 release blog post](http://blog.npmjs.org/post/161081169345/v500), it says:

> Downloads for large packages are streamed in and out of disk. npm is now able to install packages of “”“any”“” size without running out of memory. Support for publishing them is pending (due to registry limitations).

So it's possible that this, or something like it, was causing npm 3 to use too much memory and get killed, while npm 5 works okay.

The only thing that concerns me about this change is that it means our code is now running with node 8 instead of node 6. There are potentially ways to use node 6 with npm5 but I've had problems doing it on some of my own systems, and I cringe at the idea of trying to do the same on Federalist while debugging using this extremely slow feedback loop.

Since everything seems to work okay with node 8 and npm 5, I suggest we just stick with it, I guess.

msecret:
yea I agree with you, don't like the idea of using different npm / node versions

config/gulp/build.js
msecret: When just running `npm run federalist` the first npm install command would work, but then any gulp commands after that would break because gulp would switch the current directory and would try and load config from the docs folder rather then the uswds folder. So it would fail because it couldn't find `gulp release`

toolness:
Ok cool--as you suggested on Slack, we should probably add the `{% fractal_component_code %}` tag in a separate PR. Want to merge this now?

PR #358, opened 2017-07-19
Find duplicate "id" attributes when crawling site

toolness:
This is related to #357 but has more general use: we want to know if any of our pages have distinct elements with duplicate `id` attributes, because `id`s are supposed to be unique. Since `id`s are often used with JS and/or as link targets, it's particularly important that they be unique, or else the site may misbehave or take the user somewhere they don't want to go.

This PR adds support for finding duplicate `id` attributes when crawling our site:

duplicate-ids

To do
- [x] Fix the duplicate `id` errors so we don't break the build when merging this PR.

### DIFF
--- a/_data/implementations.yml
+++ b/_data/implementations.yml
@@ -7,6 +7,7 @@
   notes: "This package provides access to the Standards in [Django](https://www.djangoproject.com/) applications."
 
 - name: Django
+  id: django-uswds-forms
   url: http://django-uswds-forms.readthedocs.io/en/latest/
   author:
     name: 18F
@@ -80,6 +81,7 @@
   notes: "This is not a reusable WordPress theme, but an implementation of the Standards for [bbg.gov](https://www.bbg.gov/)."
 
 - name: WordPress
+  id: wordpress-benjamin
   url: https://github.com/kyle-jennings/Benjamin
   author:
     name: Kyle Jennings
--- a/config/crawl.js
+++ b/config/crawl.js
@@ -1,5 +1,6 @@
 const path = require('path');
 const fs = require('fs');
+const cheerio = require('cheerio');
 const express = require('express');
 const Crawler = require("simplecrawler");
 const chalk = require('chalk');
@@ -33,10 +34,43 @@ function shouldFetch(item, referrerItem) {
   return true;
 }
 
+function findDuplicateIds($) {
+  const duplicates = [];
+  const idCounts = {};
+  const ids = $("[id]").each(function() {
+    const id = $(this).attr('id');
+
+    if (!(id in idCounts)) {
+      idCounts[id] = 0;
+    }
+    idCounts[id]++;
+
+    if (idCounts[id] === 2) {
+      duplicates.push(id);
+    }
+  });
+
+  return duplicates;
+}
+
 runServer().then(server => {
   const crawler = new Crawler(`${server.url}/`);
+  const origDiscoverResources = crawler.discoverResources;
   const referrers = {};
   const notFound = [];
+  const duplicateIds = [];
+
+  crawler.discoverResources = function(buffer, item) {
+    if (/^text\/html/.test(item.stateData.contentType)) {
+      const $ = cheerio.load(buffer.toString("utf8"));
+      const ids = findDuplicateIds($);
+
+      if (ids.length) {
+        duplicateIds.push({ item, ids });
+      }
+    }
+    return origDiscoverResources.apply(this, arguments);
+  };
 
   crawler.addFetchCondition((item, referrerItem, cb) => {
     cb(null, shouldFetch(item, referrerItem));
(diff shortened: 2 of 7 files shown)


toolness:
Ok, most of the duplicate ids were due to the following kind of problem:

```md
## Solution one
### Pros
...
### Cons
...
## Solution two
### Pros
...
### Cons
...
```

The problem here is that due to https://github.com/vmg/redcarpet/issues/307, Redcarpet's `with_toc_data` extension is going to give each "Pros" and "Cons" header the same `id`.  There's a workaround proposed at https://github.com/vmg/redcarpet/issues/307#issuecomment-261793668 but it raised additional problems for me, so I ended up just converting the offending headers to raw HTML as a way to circumvent the problem. I've explained my strategy in https://github.com/vmg/redcarpet/issues/307#issuecomment-316685746.

msecret:
@toolness I actually think it seems confusing to require only some headers to be in HTML based whether it's repeated or not. I was also confused, there was mention of a more comprehensive fix, but that would require us forking redcarpet?

toolness:
Yeah, it is kinda suboptimal :(  The other option doesn't require forking redcarpet, it just requires subclassing redcarpet. Here's a diff from a git stash I made while exploring:

<details>
<summary>Diff of example Redcarpet subclass</summary>

```diff
diff --git a/_config.yml b/_config.yml
index 71fb614..849ab48 100644
--- a/_config.yml
+++ b/_config.yml
@@ -61,7 +61,7 @@ defaults:
       layout: post
 
 # Build settings
-markdown: redcarpet
+markdown: RedcarpetExtender
 redcarpet:
   extensions:
     - smart
diff --git a/_plugins/redcarpet_extender.rb b/_plugins/redcarpet_extender.rb
new file mode 100644
index 0000000..375fb9f
--- /dev/null
+++ b/_plugins/redcarpet_extender.rb
@@ -0,0 +1,52 @@
+require 'redcarpet'
+
+class RedcarpetExtender < Redcarpet::Render::HTML
+  def initialize(opts = {})
+    super
+    @anchors = []
+  end
+
+  # This resolves duplicate header IDs the "numeric" way: if you have the
+  # same title twice then a number is appended to the anchor.
+  #
+  # Taken from https://github.com/vmg/redcarpet/issues/307.
+  def header(text, level)
+    number = 2
+
+    # https://stackoverflow.com/a/4308399
+    anchor = text.downcase.strip.gsub(' ', '-').gsub(/[^\w-]/, '')
+
+    while @anchors.include?(anchor)
+      if number > 2
+        anchor[-1] = number.to_s
+      else
+        anchor += "-#{number}"
+      end
+
+      number += 1
+    end
+
+    @anchors << anchor
+
+    %(<h#{level} id="#{anchor}">#{text}</h#{level}>)
+  end
+end
+
+# https://stackoverflow.com/a/24760839
+class Jekyll::Converters::Markdown::RedcarpetExtender
+  def initialize(config)
+    @site_config = config
+  end
+
+  def extensions
+    Hash[ *@site_config['redcarpet']['extensions'].map {|e| [e.to_sym, true]}.flatten ]
+  end
+
+  def markdown
+    @markdown ||= Redcarpet::Markdown.new(RedcarpetExtender, extensions)
+  end
+
+  def convert(content)
+    markdown.render(content)
+  end
+end
```

</details>

One thing I'm concerned about with this option is that by providing our own custom function to create unique IDs for each heading, we're completely bypassing Redcarpet's default mechanism for doing so, which means that we might end up with ids that are ever-so-slightly different from the ones we've got now (we could dig into [`rndr_header_anchor()`](https://github.com/vmg/redcarpet/blob/26c80f05e774b31cd01255b0fa62e883ac185bf3/ext/redcarpet/html.c#L273-L319) to learn more). That could potentially break some permalinks, but I guess breaking anchor permalinks isn't that bad since folks wouldn't be getting a 404, they just wouldn't be sent to the exact place on a page they're looking for.

That said, the main reason I didn't end up pursuing this solution was because the Redcarpet subclass was somehow causing relative URLs in our generated CSS to be incorrect. I had no idea why this was happening, and combined with my unease about breaking existing ids, I decided to use the raw HTML strategy used in this PR.

I could try going back to the Redcarpet subclassing solution and figuring out exactly why it was mangling the CSS like that. I should probably timebox my investigation. though.

toolness:
Hmm, so it turns out that the Stack Overflow thread I had gotten the "how do I to use a custom Redcarpet renderer with Jekyll" answer from was kind of wrong, or at least out-of-date. Based on my reading of Jekyll's [`redcarpet_parser.rb`](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/converters/markdown/redcarpet_parser.rb), it seems like we should either be subclassing `Jekyll::Converters::Markdown::RedcarpetParser` and overriding its `class_with_proper_highlighter` method, or perhaps just monkeypatching `Redcarpet::Render::HTML` itself so that all instances of the default parser use our custom header rendering logic. I'm a little bit concerned about how stable such a solution will be as Jekyll continues to evolve, though... Blerg. I wish Jekyll made it a bit easier (and more well-documented) to change the renderer it used.

toolness:
Closing this in favor of #363!

PR #363, opened 2017-07-24
Find duplicate "id" attributes when crawling site (Redcarpet subclassing approach)

toolness:
This is just like #358 but takes an alternative approach to fixing duplicate ids caused by the kinds of header TOC issues described in https://github.com/18F/web-design-standards-docs/pull/358#issuecomment-316686935. Instead of using HTML in the markdown for duplicate headers, we subclass Redcarpet to provide our own custom id-collision algorithm.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -61,7 +61,7 @@ defaults:
       layout: post
 
 # Build settings
-markdown: redcarpet
+markdown: RedcarpetExtender
 redcarpet:
   extensions:
     - smart
--- a/_data/implementations.yml
+++ b/_data/implementations.yml
@@ -7,6 +7,7 @@
   notes: "This package provides access to the Standards in [Django](https://www.djangoproject.com/) applications."
 
 - name: Django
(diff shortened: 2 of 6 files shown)


toolness:
One thing we might want to consider with this approach is removing the `with_toc_data` extension from our `_config.yml`. If all that extension does is add `id` attributes to generated headers, then we don't need it anymore--but if it's doing anything extra then we might want to keep it around.

toolness:
Cool, sounds good!

This will not help with the id problems from the fractal components because that's HTML, which the markdown parser just leaves as-is.  Our custom id logic specifically only gets triggered when Redcarpet is converting Markdown-style headers like `## Section title` into HTML.

toolness:
Er wait, were there actual changes you wanted me to make in this PR?  It says you "requested changes" but I'm not seeing any information about changes to make...

msecret:
@toolness Don't think I meant to request changes, sorry.

PR #366, opened 2017-07-25
initial fractal link on Buttons component

donjo:
screen shot 2017-07-25 at 1 53 46 pm

Adds a "view in component library" link to the pages on the docs site (just buttons for now)

@toolness is there anyway I could use the code you wrote for embedding fractal code samples to populate that URL field?

### DIFF
--- a/_components/buttons.md
+++ b/_components/buttons.md
@@ -7,7 +7,7 @@ type: element
 title: Buttons
 category: UI components
 lead: Use buttons to signal actions.
-component_url: 'https://components.standards.usa.gov/components/detail/buttons.html'
+fractal_component_url: 'https://components.standards.usa.gov/components/detail/buttons.html'
 ---
 
 {% include code/preview.html component="buttons" %}
--- a/_components/buttons.md
+++ b/_components/buttons.md
@@ -7,6 +7,7 @@ type: element
 title: Buttons
 category: UI components
 lead: Use buttons to signal actions.
+component_url: 'https://components.standards.usa.gov/components/detail/buttons.html'
 ---
 
 {% include code/preview.html component="buttons" %}
--- a/_layouts/styleguide.html
+++ b/_layouts/styleguide.html
@@ -16,13 +16,15 @@ layout: default
     <header>
       {% if page.category != null %}
         {% unless page.path contains 'overview' %}
-      <p class="site-subheading">{{ page.category }}</p>
+          <p class="site-subheading">{{ page.category }}</p>
         {% endunless %}
       {% endif %}
       <h1 id="{{ page.title | slugify }}">{{ page.title }}</h1>
-      <span class="components-library-link">
-        <a href="{{ page.component_url }}">View in Component Library</a>
-      </span>
+      {% if page.category == 'UI components' and page.component_url %}
+        <span class="components-library-link">
+          <a href="{{ page.component_url }}">View in Component Library</a>
+        </span>
+      {% endif %}
     </header>
 
     {% include lead.html text=page.lead %}
--- a/_layouts/styleguide.html
+++ b/_layouts/styleguide.html
@@ -20,7 +20,7 @@ layout: default
         {% endunless %}
       {% endif %}
       <h1 id="{{ page.title | slugify }}">{{ page.title }}</h1>
-      {% if page.category == 'UI components' and page.component_url %}
+      {% if page.category == 'UI components' and page.fractal_component_url %}
         <span class="components-library-link">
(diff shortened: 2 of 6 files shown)


toolness:
> @toolness is there anyway I could use the code you wrote for embedding fractal code samples to populate that URL field?

Hmm, so we're never actually referring to the *details* page for components in the fractal code samples section, only the *raw* HTML page, so there's not a ton of DRY savings we can make there, since the URLs are actually different. But there is one easy way I can think of to make these component pages less verbose, and that's to change the `component_url` front matter key to just `fractal_component`, and then put the vast majority of that URL in the `_layouts/styleguide.html` template, since most of it never changes.

Once you've merged this, and once #357 is merged, I can add a `{% link_to_fractal_component 'buttons' %}` tag that returns an absolute URL for the given component *and* checks the filesystem to make sure that component actually exists, throwing an exception if it doesn't.  That way we can at least make sure that we're never accidentally linking to components that don't exist.

I can also make this tag take advantage of the `FRACTAL_BASE_URL` environment variable (introduced in #357) so that if it exists, the link will point to the developer's own local fractal instance instead of standards.usa.gov.  This will make it easier for us to locally develop!

_layouts/styleguide.html
```diff
@@ -20,6 +20,9 @@
         {% endunless %}
       {% endif %}
       <h1 id="{{ page.title | slugify }}">{{ page.title }}</h1>
+      <span class="components-library-link">
+        <a href="{{ page.component_url }}">View in Component Library</a>
```
toolness: So based on my suggestion in https://github.com/18F/web-design-standards-docs/pull/366#issuecomment-317887508, you could change this line to:

```html
<a href="https://components.standards.usa.gov/components/detail/{{ page.fractal_component }}">View in Component Library</a>
```

That way individual components just have to define e.g. `fractal_component: buttons` in their front matter instead of a big 'ol URL.

Oh! I just noticed you've also defined a `components_url` in your `_config.yml`, so maybe the above HTML could even be something like:

```html
<a href="{{ site.components_url }}/components/detail/{{ page.fractal_component }}">View in Component Library</a>
```

That way if we eventually decide to have e.g. a staging site for the fractal components + docs, we can have them link to each other, rather than the docs always hard-linking themselves to the production fractal site.

toolness:
This is looking good!

One recommendation: I think variables/settings you're introducing that have the word "component" in them (e.g. `components_url`) should be qualified to have the word "fractal" in them too (e.g. `fractal_components_url`), just to be super explicit.

Otherwise I think that the naming is ambiguous, since the docs site itself has its own "UI components" section. For instance, if someone sees some code that mentions `components_url`, they may assume it's referring to https://standards.usa.gov/components/, rather than the components site, which would be confusing.

donjo:
@toolness can you take another look, I made some modifications and added some logic. Is this everything we should do before #357?

_layouts/styleguide.html
```diff
@@ -16,13 +16,15 @@
...
         {% unless page.path contains 'overview' %}
-      <p class="site-subheading">{{ page.category }}</p>
+          <p class="site-subheading">{{ page.category }}</p>
         {% endunless %}
       {% endif %}
       <h1 id="{{ page.title | slugify }}">{{ page.title }}</h1>
-      <span class="components-library-link">
-        <a href="{{ page.component_url }}">View in Component Library</a>
-      </span>
+      {% if page.category == 'UI components' and page.fractal_component_url %}
```
toolness: Oh interesting, it's not enough to just test if `page.fractal_component_url` is truthy here? What's the situation where it isn't enough? I recall you mentioning on Slack that you didn't realize that this layout was used on non-components pages, but I would've thought that if it was used elsewhere, `page.fractal_component_url` would be falsy...

_components/buttons.md
```diff
@@ -7,7 +7,7 @@ type: element
 title: Buttons
 category: UI components
 lead: Use buttons to signal actions.
-component_url: 'https://components.standards.usa.gov/components/detail/buttons.html'
+fractal_component_url: 'https://components.standards.usa.gov/components/detail/buttons.html'
```
toolness: One way to make this both less verbose for other components and easier to transition over to making the base fractal URL being supplied by an environment variable would be to set `fractal_component_url: '/components/detail/buttons.html`, and move the `https://components.standards.usa.gov` part of the string into `styleguide.html`, so that it's provided by the template rather than the YAML front matter. It's up to you though, I guess we can also just change it after #357 lands!

PR #367, opened 2017-07-26
Add links to graphic list on landing page

timwis:
## Description

The landing page has a graphic list resembling the site navigation, which I went to click on, but noticed they're not links. I added conditional `<a>` tags to the template, trying to reuse the property name `href` from other YAML in this repo, but I noticed it adds some padding to the top, making a more significant change than I expected. As a result, this pull request is not quite merge-ready, but hopefully it's a helpful start! Didn't want to take too many liberties in adjusting the padding - only got my toes wet :)

## Additional information

screen shot 2017-07-26 at 11 22 14
^ An example of the vertical padding difference when compared with a graphic list item that doesn't have an `href`.

### DIFF
--- a/_layouts/landing.html
+++ b/_layouts/landing.html
@@ -28,9 +28,23 @@ layout: default
     <div class="usa-grid usa-graphic_list-row">
       {% for list_item in page.graphic_list.list_item %}
         <div class="usa-width-one-fourth usa-media_block">
-          <img class="usa-media_block-img"  src="{{ list_item.image.src | relative_url }}" alt="{{ list_item.image.alt }}">
+          {% if list_item.href %}
+            <a href="{{ list_item.href | relative_url }}">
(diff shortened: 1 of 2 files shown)


donjo:
@timwis Hey Tim, thanks for pointing this out. We had talked about doing a card style for these similar to how the "Featured Projects" section on the [18F](https://18f.gsa.gov/) site works where the user could click anywhere in the box to be taken to the appropriate section (the link, text, and header would all be one active link). I was tentatively scheduling the design/implementation of it for next sprint (starting Monday) but since you started this PR I wanted to reach out and see if you think a solution like that would solve the problem. What do you think?

timwis:
It totally would, I like that much better :) feel free to close this or just pull in the YAML changes when appropriate.

donjo:
@timwis Thanks Tim, I'll close this out and work on it in a new issue.

PR #373, opened 2017-08-03
[Website] Nicer error message/behaviour if no configuration.

dracos:
Rather than `Error:  undefined method '[]' for nil:NilClass` or `Error:  undefined method 'each' for nil:NilClass`, print out a message saying the config is missing, or allow through an empty config.

### DIFF
--- a/_plugins/incremental_regeneration_fixer.rb
+++ b/_plugins/incremental_regeneration_fixer.rb
@@ -14,8 +14,17 @@ module IncrementalRegenerationFixer
     interdependent_files = []
 
     config = site.config['incremental_regeneration_fixer']
+    if not config
+      raise ("You probably want to specify incremental_regeneration_fixer " +
+             "config if you want to use this plugin")
+    end
 
-    for glob in config['interdependent_files']
+    globs = config['interdependent_files']
+    if not globs
+      return
+    end
+
+    for glob in globs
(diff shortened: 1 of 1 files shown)


AndreaSigz:
Thanks @dracos @donjo @toolness

dracos:
@toolness Thanks :) I had found https://github.com/18F/web-design-standards-docs/pull/334 after a relook at the related 4112 issue I had commented on previously; I'd copied the file into _plugins but hadn't yet put any config in the YAML (distracted part way through probably), so got the error. Then later I wanted to be able to comment out the config without also having to remove the file (to test different settings).

toolness:
Interesting! Are you currently using a fork of this repository to generate your own custom style guide or something? I'm just curious how it's being used outside of standards.usa.gov is all, no worries if you'd rather not say.

dracos:
Sorry, no, this was for http://fixmystreet.org/ by @mySociety – I just like developing it with `--incremental` and it was breaking our blog and I found your plugin workaround :)

toolness:
Oh @mysociety rocks!  That's cool, I had just made that plugin as a hack since I like `--incremental` too, but if it's more broadly useful, I guess I could pull it out into its own repository...

Anyways, thanks for the contribution and the explanation!

PR #378, opened 2017-08-17
Disable form submission in HTML for sensitive forms

toolness:
This is an attempt to fix https://github.com/18F/web-design-standards/issues/2101.

As I understand it, the main rationale for disabling form submission, at least in this specific instance, is so that security scans on standards.usa.gov don't trigger false positives.  I wasn't forwarded the original email from the GSA IT folks though, so I'm not sure.

The thing is, #202 disables form submission for all components from the client-side, and I've verified that code is still working fine. However, if JS is disabled, it won't take effect. My guess is that the security scanner is trying to make sure that the site is secure even with JS disabled, so it's essentially evaluating the page that way.

I *think* one solution is to disable form submission at the HTML level, which is what this PR does. It seems to work when I disable JS on my browser.

The main downside with this approach, though, is that the `disabled` attribute on the submit buttons shows up in the "Code" accordion.  That's a bit annoying I guess, but it's not a deal-breaker--the alternative would be to have some sort of liquid tag that automatically inserts `disabled` attributes as needed before passing the HTML on to the syntax highlighter, but that's a lot of work.

### DIFF
--- a/_includes/code/components/password-reset.html
+++ b/_includes/code/components/password-reset.html
@@ -33,6 +33,6 @@
         Show my typing</a>
     </p>
 
-    <input type="submit" value="Reset password">
+    <input type="submit" value="Reset password" disabled>
   </fieldset>
 </form>
--- a/_includes/code/components/sign-in-form.html
+++ b/_includes/code/components/sign-in-form.html
@@ -15,7 +15,7 @@
         Show password</a>
     </p>
 
-    <input type="submit" value="Sign in">
+    <input type="submit" value="Sign in" disabled>
 
     <p><a href="javascript:void(0);" title="Forgot username">
       Forgot username?</a></p>


msecret:
I'm not sure where I saw this, but I think somewhere previously, we had some DOM code that disabled all forms on the page. Something like:

```js
$('form').on('submit', function(ev) { ev.preventDefault(); });
```

I'll try and find it. But I think somebody was saying something like that should be added to the components site.

wslack:
Marco's solution sounds right. I'll forward the e-mail to Atul.

msecret:
Found it: https://github.com/18F/web-design-standards-docs/blob/develop/js/form-disable.js

wslack:
The vibe I got is that the scanning software wants to make sure that people won't accidentally send a username/pw in a URL.

toolness:
@msecret yep, that's the code I was referring to in the description of this PR--it was added in #202.  As I mentioned there, I think the scanning software is trying to ensure that the scanning software is specifically trying to make sure people won't accidentally send the username/pw in a URL when JS is disabled on their end, which #202 doesn't account for.

toolness:
(The less charitable explanation is that the security scanner doesn't run the JS on the website because it's not "advanced enough", but since this still effectively emulates a user for whom JS won't work, either because of network error or b/c their browser has it disabled, it seems like a reasonable concern.)

msecret:
Did the scanning software just fail for the components site, or for the standards.gov site too?

toolness:
Whoa, I had not even considered the possibility that the scanning software was scanning components.standards.usa.gov. But now that I read the description of 18F/web-design-standards#2101, it is specifically saying "components site", so that is my bad. I had assumed it was talking about the components *section* of the standards.gov site, which is why I wrote up this PR.

I'm now assuming I had just read the issue wrong, so I'll close this PR!

wslack:
@toolness I'm sorry - yeah, standards.usa.gov is already live on the Federalist SSP. This is for the Fractal site!

Issue #386, opened 2017-08-22
Create a RELEASE.md file


toolness:
We should store instructions on how to cut a new release in this repository, rather than inlining it into the core USWDS's `RELEASE.md`, as mentioned in https://github.com/18F/web-design-standards/issues/2124.

PR #387, opened 2017-08-23
Add instructions on how to update the USWDS version

toolness:
This adds instructions on how to update the USWDS version to `README.md`.

It's not quite a `RELEASE.md` (#386) but it helps get us there.  I figure it's also useful to put in the `README.md` since devs will want to know how to do it outside of the context of a release, e.g. to get the latest fractal component HTML from a specific USWDS commit.

When writing the docs and trying out the `npm install` commands, I noticed that npm prefixes the dependency name with the `github:` protocol, and it also seems to lowercase the organization name. So I updated `_plugins/uswds_version.rb` to support that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,30 @@ Federalist also builds public previews for each branch pushed to GitHub. For ins
 
 https://federalist.fr.cloud.gov/preview/18f/web-design-standards-docs/develop/
 
+### Updating the USWDS version
+
+To update the version of USWDS being used, change the version that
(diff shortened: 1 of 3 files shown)


donjo:
I forget why this has been sitting here for a while but now that I was ready to review it I realized some things will need to be updated to account for the new repo name/location before it's ready to go.

maya:
@donjo did you want to make those updates or should I?

maya:
Updated the paths in 2dcbc6f.

PR #388, opened 2017-08-24
Release USWDS 1.3.1

### LINKED ISSUE #385: Release a new version that targets USWDS 1.3.1
We just [released USWDS v1.3.1](https://github.com/18F/web-design-standards/releases/tag/v1.3.1) so we should list it on the production docs site!

toolness:
This updates the version of uswds the `master` branch uses to 1.3.1, which was just released in https://github.com/18F/web-design-standards/pull/2119, so that we provide the download link for that version instead of the old one.

Fixes #385.

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
     "politespace": "^0.1.4",
     "run-sequence": "^1.1.5",
     "simplecrawler": "^1.1.3",
-    "uswds": "1.3.0",
+    "uswds": "1.3.1",
     "vinyl-buffer": "^1.0.0",
     "vinyl-source-stream": "^1.1.0"
   }


toolness:
Hmm, it looks like CircleCI is failing:

```
  Conversion error: Jekyll::Converters::Scss encountered an error while converting 'css/styleguide.scss':
                    File to import not found or unreadable: lib/bourbon. Load paths: /home/ubuntu/web-design-standards-docs/css /home/ubuntu/web-design-standards-docs/assets/css on line 4
```

This is weird because I don't understand why this is failing, where #361 wasn't.

Any ideas @donjo ?

toolness:
wut? i didn't restart CircleCI myself, but it somehow restarted, and everything is fine now... bizarre!

i wonder if it was merging against `develop` before or something?

whatever, i am confused and circleCI seems to be happy now.

PR #400, opened 2017-10-02
update sidenav IDs to stay consistent with other examples

donjo:
fixes 18F/web-design-standards#2150


donjo:
closing this out since the code changed in #401 and is now being imported via fractal components

PR #401, opened 2017-10-03
switching header to fractal template tags

donjo:
This switches the header sections of the docs site to use code coming from Fractal for both the code samples and the generated examples on the demo pages. 

As a side bonus, it fixes the problem where the government banner was out of date on the header examples.


donjo:
FYI this doesn't fix the header issue (that's in `web-design-standards`) but you can confirm this is working by looking at the generated code samples and seeing that they have the updated Gov Banner text and the links in the header are derived from [this](https://github.com/18F/web-design-standards/blob/develop/src/components/08-navigation/nav-primary.config.yml) navigation config file.

maya:
It's showing up now on the desktop view, but I see this in mobile:
screen shot 2017-10-03 at 3 46 14 pm

which goes to this when it's smaller:
screen shot 2017-10-03 at 3 48 11 pm

Quick visual design check, the header now does not extend to the width of the page like before. I checked @rtwell's designs in https://github.com/18F/web-design-standards/issues/2086 and it seems to be full-page width too, not cropped. Was this change intended or decided somewhere else? 

### Current

screen shot 2017-10-03 at 3 41 35 pm

### Design via https://github.com/18F/web-design-standards/issues/2086 

screen shot 2017-10-03 at 3 42 21 pm

/cc @thisisdano

Issue #371, opened 2017-07-31
update password requirements example


donjo:
make the reset password box use sample text that doesn't look like a real password requirement

maya:
Fixed in #2160.

PR #403, opened 2017-10-04
converting password reset to fractal template

donjo:
Goes with the fix of #371 in 18f/web-design-standards#2160


maya:
@donjo can we also add a note in the documentation about the NIST guidance per https://github.com/18F/web-design-standards/issues/1186?

PR #427, opened 2017-10-27
[Grid]: Add code samples to grids page

### LINKED ISSUE #356: Grids page lacks example code
As of https://github.com/18F/web-design-standards/pull/2038, we now have docs and code examples for the grid component on the fractal site, but we don't have code examples on this docs site.

A funny thing I discovered is that this is actually a regression: I was watching the old USWDS developer-focused tour of the standards from November 2015 and [at 32:22](https://youtu.be/eMIWzj_dPvg?t=32m22s) @maya shows off the grid page, and the example code for the grid is right there. So it got removed at some point.

I tried briefly re-adding the current example as a code component preview/accordion pair, as most of the other code examples are currently done, but I found one wrinkle, which is probably why this component never got the same treatment: the code examples need to be wrapped in a `grid-example` div, and that class is specific to the style guide, as it is used to "illustrate invisible grid for web design standards", as the comment next to it says. Without that special class, the example actually looks quite janky. But it also feels wrong to use it as an example that folks can just copy-paste, when it uses a CSS class specific to the style guide, so I just left it as-is.

maya:
There are no code samples on the grid page anymore, this adds it back in.

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/18f/web-design-standards-docs/fix-grid-code-sample/components/grids/)

Fixes #356.

### DIFF
--- a/_components/grids.md
+++ b/_components/grids.md
@@ -8,123 +8,8 @@ category: UI components
 lead: This 12-column, responsive grid provides structure for website content.
 ---
 
-<h2 class="usa-heading">Grid</h2>
-<div class="preview preview-no_border grid-example grid-example-blank">
-
-  <div class="usa-grid">
-    <div class="usa-width-one-whole">1/1</div>
-  </div>
-  <div class="usa-grid">
-    <div class="usa-width-one-half">1/2</div>
-    <div class="usa-width-one-half">1/2</div>
-  </div>
-  <div class="usa-grid">
-    <div class="usa-width-one-third">1/3</div>
-    <div class="usa-width-one-third">1/3</div>
-    <div class="usa-width-one-third">1/3</div>
-  </div>
-  <div class="usa-grid">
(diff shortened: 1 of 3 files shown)


donjo:
There is a fractal example we could insert here thought it's styled a bit differently. Should we populate `_includes/code/components/grids.html` with a fractal_component reference instead?

maya:
@donjo Yes it probably should be done at some point. I prefer the styles on the docs site (clearer, easier to see). Not a big fan of pulling in the scoped styles from fractal into the code samples which is what it would do now. Maybe we tackle it in a separate PR?

donjo:
Sounds good to me, I'll create a new issue. One for figuring out the scoped styles and one for updating the grids component to use fractal code.

maya:
@donjo I also added grids to the list here: https://github.com/18F/web-design-standards-docs/issues/415

PR #440, opened 2017-11-06
CC 18F bizdev to inquiry email link

maya:
- Adds `+training` and `+customization` to email links for ease of filtering. See: https://gmail.googleblog.com/2008/03/2-hidden-ways-to-get-more-from-your.html
- Adds cc to 18F business development

cc @d-kenny

### DIFF
--- a/pages/home.md
+++ b/pages/home.md
@@ -40,7 +40,7 @@ product team will provide guidance for getting up and running with
 the U.S. Web Design Standards and kick-start your design and
 development.
 
-<a href="mailto:uswebdesignstandards@gsa.gov" class="usa-button cta" onclick="ga('send', 'event', 'Clicked Training CTA', 'Clicked training call to action');">
+<a href="mailto:uswebdesignstandards+training@gsa.gov?cc=Inquiries18F@gsa.gov" class="usa-button cta" onclick="ga('send', 'event', 'Clicked Training CTA', 'Clicked training call to action');">
   <img src="{{ site.baseurl }}/img/logo-email.png" alt="">
   Email us about training
 </a>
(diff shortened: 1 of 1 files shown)


maya:
Since we're adding the cc inquiries18F@gsa.gov, we can always search by that and probably don't need the `+training` so removed in 8d938ba.

thisisdano:
It was such a clever solution, I kinda miss it.

maya:
@thisisdano We can re-add it, do you like the idea of just having it say `+inquiry` for both or have it grouped out?

thisisdano:
I like it grouped out as `+training` and `+customization`

maya:
Np! Quick question: on the latest home page designs we found, "customization" changed to "consultation"... is that something we plan on doing. @donjo do you recall if 18F bizdev wanted that phrasing? 

cc @d-kenny @MKathrynC

donjo:
@maya I can't remember whether or not that was a suggestion or an ad-hoc decision. I do remember it being discussed whether or not "training" would be something that we mention/offer on the homepage but not sure where the team stands on that today so better to let the folks you cc'd chime in.

maya:
It sounds like we're removing "Training." @thisisdano @donjo any preference on the display below:

screen shot 2017-11-08 at 8 51 31 am
screen shot 2017-11-08 at 9 50 14 am
screen shot 2017-11-08 at 8 58 39 am

maya:
Re-added +training and +customization to email links in 13d16ae. 

Holding off on content changes until @d-kenny has a meeting to decide this next Weds. 

This is ready to review.

donjo:
@maya FWIW I like the third option of the new content styles when it comes to tackling that in a new issue. I'll review this so the address changes can get pulled in now

PR #443, opened 2017-11-07
Use extended header navigation styles for docs site

### LINKED ISSUE #296: Update docs site header navigation design
For the horizontal navigation bar on the site, should it be using the navigation that comes with the extended header? That would make links easier to click (larger click area) and have consistent current link styles (border bottom), and would be more consistent in general.

**USWDS docs site**:

screen shot 2017-05-15 at 5 10 50 pm

**USWDS extended header pattern**:
screen shot 2017-05-15 at 5 10 12 pm

maya:
[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/18f/web-design-standards-docs/update-header-nav)

Fixes #296.

### DIFF
--- a/_includes/nav/page-item.html
+++ b/_includes/nav/page-item.html
@@ -1,7 +1,9 @@
 {% include nav/current.html page=include.page %}
 <a href="{{ include.page.url | relative_url }}"{%
   if _current -%} class="usa-current"{% endif %}>
-  <span>{{ include.page.title | smartify }}</span>
+  {% unless sidenav %}<span>{% endunless %}
+    {{ include.page.title | smartify }}
+  {% unless sidenav %}</span>{% endunless %}
 </a>
 {% if _current -%}
   {% unless include.flat -%}
--- a/_includes/nav/page-item.html
+++ b/_includes/nav/page-item.html
@@ -1,8 +1,8 @@
 {% include nav/current.html page=include.page %}
 <a href="{{ include.page.url | relative_url }}"{%
-  if _current -%} class="usa-current"{% endif %}>{{
-    include.page.title | smartify
-  }}</a>
+  if _current -%} class="usa-current"{% endif %}>
+  <span>{{ include.page.title | smartify }}</span>
+</a>
 {% if _current -%}
   {% unless include.flat -%}
     {% assign _href_key = include.href_key %}
--- a/css/styleguide.scss
+++ b/css/styleguide.scss
@@ -273,7 +273,7 @@ body {
     clear: both;
     display: block;
     height: $height-nav-secondary;
-    padding-left: 3rem;
+    padding-left: 1.5rem;
     z-index: $z-nav-secondary;
   }
 
--- a/css/styleguide.scss
+++ b/css/styleguide.scss
@@ -269,7 +269,7 @@ body {
 
   @include media($nav-width) {
     background-color: $color-white;
-    border-bottom: 1px solid $color-gray-light;
+    border-bottom: 1px solid $color-gray-lighter;
     clear: both;
     display: block;
     height: $height-nav-secondary;
--- a/css/styleguide.scss
+++ b/css/styleguide.scss
@@ -282,25 +282,30 @@ body {
 
     li {
       display: inline;
-      padding-right: 3rem;
     }
   }
 
   a {
-    color: $color-gray-dark;
-    font-size: $site-nav-font-size;
-    font-weight: normal;
-    letter-spacing: 0.3px;
+    color: $color-gray;
+    font-size: $h5-font-size;
+    font-weight: $font-bold;
     line-height: $height-nav-secondary;
+    padding: 1.5rem;
     text-decoration: none;
+    -webkit-font-smoothing: antialiased;
 
     &:hover {
-      color: $color-primary;
-      text-decoration: underline;
+      span {
+        @include nav-border-thick;
+        color: $color-primary;
+      }
     }
+  }
 
-    &.usa-current {
-      color: $color-primary;
+  .usa-current {
+    span {
+      @include nav-border-thick;
+      color: $color-base;
     }
   }
 }


_includes/nav/page-item.html
```diff
@@ -1,8 +1,8 @@
 {% include nav/current.html page=include.page %}
 <a href="{{ include.page.url | relative_url }}"{%
-  if _current -%} class="usa-current"{% endif %}>{{
-    include.page.title | smartify
```
maya: _Re-adding in this new PR_ 
@donjo do you know how I can add the `<span>` to the header nav but not the sidenav?

_includes/nav/page-item.html
```diff
@@ -1,8 +1,8 @@
...
-  }}</a>
+  if _current -%} class="usa-current"{% endif %}>
+  <span>{{ include.page.title | smartify }}</span>
```
maya: _Re-adding in this new PR + moving to the correct line of code_ 
@donjo do you know how I can add the `<span>` to the header nav but not the sidenav?

donjo: Ok in trying to figure out how all of this is sew together and I think I now understand. Which part are you trying to ensure doesn't happen on the sidenav? just the "current" styling that adds the blue bar on the left or are you trying to omit something from the sidenav?

maya: I just don't want the `<span>` wrapped around the title in the sidenav but do want it in the main nav. Lmk if that makes sense.

Header nav: `<a href=""><span>Title</span></a>`
Sidenav: `<a href="">Title></a>`

There are actually no visual changes from this, it's just the markup.

donjo: Ok that makes sense. 

🤔  _thinking_

donjo: not sure if there is a more elegant way to write this but this works

```
<a href="{{ include.page.url | relative_url }}"{%
  if _current -%} class="usa-current"{% endif %}>
  {% unless sidenav %}<span>{% endunless %}
  {{ include.page.title | smartify }}
  {% unless sidenav %}</span>{% endunless %}
</a>
```

maya: Perfect 🎉 Added in 3942ae6.

maya:
@donjo or @thisisdano this is ready for review.

thisisdano:
1) This looks perfect.

2) I feel silly asking this, but if this is the default extended header nav why does it need custom styling  — are there things that our USWDS styles aren't catching (or can't catch)? Is it because we've diverged from the default extended header? Is it a problem that we don't build our own header consistent with our guidance and header component?

donjo:
While I'm all for using the code of the extended header and inheriting the nice underlines that it provides. I'm not a fan of the changes to the text in the navbar. I find the smaller, bolder, and lighter text harder to read and I don't feel like it fits as well with the typography we use on the rest of the homepage. This change brings it closer to how it once around v1.0 was but strays from the cleaner look we were going for with the interim homepage re-design a few months ago. 

New:
screen shot 2017-11-08 at 2 02 54 pm

Old:
screen shot 2017-11-08 at 2 02 32 pm

maya:
@thisisdano re: #2 
1.) The docs site header was built way before our USWDS headers.
2.) They are functionally different. The USWDS headers has a single HTML markup that's transformed from a mobile nav to the desktop header nav. While the docs site has different content and markup for the mobile nav and desktop header nav, so are hiding and showing them at different sizes. The docs site nav has its child pages represented in the sidenav, whereas the USWDS nav would have the child pages in a dropdown menu within the main nav. There's a lot about the docs site that is unique and not part of the Standards. Definitely something to revisit at some point.

maya:
Style-wise, I think it's good to use our default styles bc it's what all of our users use. I don't find it hard to read, it's probably good that it's visually distinct from the sidenav bc it makes the hierarchy more apparent.

thisisdano:
So, squaring the circle here, it seems like there are plenty of open questions about how much the USWDS site is an expression of the Standards themselves. 

To @donjo's point: I agree that the default navbar is less graceful and possibly less legible than the custom navbar currently implemented on the site. 

To @maya's points: I see that our site has grown and matured at a different rate than the Standards themselves, often using patterns we don't include in the Standards — partly because of necessity, and partly as a kind of proving ground (or USWDS Zen Garden) for what the Standards _could_ be.

I think it's a problem that we can't (or, in some instances, just don't) build the Standards site from the basic Standards kit, with minimal variance. If we want to build the capability for variance into the system, we should think about methods more flexible and future-forward than custom CSS. And while the Standards can be "skinned", I believe (pretty strongly) that we should be building and styling default components in such a way that they are sturdy enough and even `:blush:` _beautiful_ enough to use ourselves, and use successfully.

They aren't there yet.

But if we want to know how the Standards perform for folks who build with them, we should live with them more intimately. And if (and this really isn't an _if_ but a _when_), by doing so, we learn that a component could be improved, we should work to add that improvement to the default styles we promote.

So, I suggest we regress aesthetically in the short term so that we can move forward in the longer term. And we should continue to identify places on our own site where we could use a default style _and_ work to improve that style for everyone.

So, to me this begs the question: Can our default style for the extended nav be improved? Or, if it's important that we have both bold and non-bold variants, is this something that could be controlled with variables rather than custom CSS?

**In this instance:** I suggest we move forward and make our nav the default nav, even though it's _possibly_ a regression, then use subsequent issues to address what we might do to improve the nav for everyone.

maya:
@thisisdano can you clarify what you want us to do in terms of this PR? It's still not clear to me.

maya:
Following up on this, I still think it's a good idea to merge this.

Here’s why: we get more usable links bc they take up the full area, nice current page states, and get to eat our own dog food by using our own header designs.

We can always tweak the typography if we don’t like it @donjo, and hey, that could help inform the USWDS header too.

And yes @dano, I absolutely want us to use our code, but bc of all the functionality differences between our site and the USWDS header, think we’d be better served to do this in the future when we have more time to figure it out, and focus on delivering value that takes less effort right now.

thisisdano:
Good work on this, everyone — I'm happy to see how this issue progressed and all the thoughtful feedback all the way through

PR #444, opened 2017-11-07
Remove version from implementations

### LINKED ISSUE #351: Remove version number from list of implementations?
I noticed that we report the version number in our list of [implementations](https://standards.usa.gov/getting-started/implementations/), but I'm gonna question this because:

* It's almost guaranteed to be inaccurate. Software evolves at a rapid pace, and unless we use some sort of API, us manually updating the `version` field in `implementations.yml` is unlikely to keep up with it.

* I'm unclear on the immediate user need it serves. If the reader really wants to know what the latest version is, can't they just click on the link and see the 100% accurate version number on the project's website?

maya:
Fixes #351.

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/18f/web-design-standards-docs/remove-version-implementations/getting-started/implementations/)

### DIFF
--- a/pages/getting-started/implementations.md
+++ b/pages/getting-started/implementations.md
@@ -38,7 +38,6 @@ If you have a new implementation to add to this list, please [open an issue] or 
   <thead>
     <tr>
       <th>Distribution</th>
-      <th>Version</th>
       <th>Author or maintainer</th>
       <th>Notes</th>
     </tr>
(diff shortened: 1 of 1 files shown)


thisisdano:
This is a bit of a confusing one, since `version` seems to want to link to the USWDS version, but does not in most cases. And "version" doesn't make it clear that it's referring to the USWDS version in the first place. 

**Is there _any_ value in communicating the USWDS version that the implementation uses?** There _might_ be? I feel like I'm happy to 💣 the column, but just want to make sure that since this is referencing the USWDS version and not the 3rd-party project version, we're all on the same page that the column should be removed.

cc @toolness

maya:
For maintenance reasons, having a list we have to manually update doesn't sounds like something we want to do bc we have no way of knowing when something changes. However, linking directly allows folks to find out the version accurately.

thisisdano:
Works for me. :shipit:

PR #450, opened 2017-11-09
Update template code to use fractal

donjo:
This updates the template code to use fractal components.

Note: Currently the page that generates the demo site under `_components/form-templates/form-templates.md` isn't using fractal code because of a problem with path to the CSS. Trying to come up with a way around that. 

fixes #452


donjo:
Currently blocked until we figure out a way to not run aXe tests on these pages. Because assets live in different places in the fractal site and the docs site, a piece of the code coming from fractal returns a 404 on the docs site.

maya:
Should we move it to icebox or backlog in the meantime?

donjo:
moved to backlog for now

maya:
Superceded by #572.

PR #458, opened 2017-12-06
Circle Update

donjo:
This updates the CircleCI config to use circle 2.0 to speed up build times and unlock some new features.

In a future issue we will look into breaking out some of the tasks into individual steps so they can be run in parallel for even faster builds.


maya:
Noting that this speeds build times from 6+ mins to 4+ mins 🎉

circle.yml
```diff
@@ -1,23 +0,0 @@
...
-    # https://discuss.circleci.com/t/using-the-latest-version-of-chrome-on-circleci/871/4
-    - curl -L -o google-chrome.deb https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
-    - sudo dpkg -i google-chrome.deb
-    - sudo sed -i 's|HERE/chrome\"|HERE/chrome\" --disable-setuid-sandbox|g' /opt/google/chrome/google-chrome
-    - rm google-chrome.deb
-
-test:
-  pre:
-    - gulp -v
-    - scss-lint -v
```
toolness: Huh, was the old CircleCI config actually running `scss-lint` _twice_, once here and once as part of `npm test` below at line 21?

donjo: yeah 😢

PR #462, opened 2017-12-18
Upgrade USWDS version to 1.4.4

donjo:
Upgrades the USWDS package to 1.4.4

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "politespace": "^0.1.4",
     "run-sequence": "^1.1.5",
     "simplecrawler": "^1.1.3",
-    "uswds": "1.4.3",
+    "uswds": "^1.4.4",
     "vinyl-buffer": "^1.0.0",
     "vinyl-source-stream": "^1.1.0"
   }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -93,7 +93,7 @@
     "politespace": "^0.1.4",
     "run-sequence": "^1.1.5",
     "simplecrawler": "^1.1.3",
-    "uswds": "1.4.3",
+    "uswds": "^1.4.4",
```
maya: Carrot needs to be removed. We usually add this manually, so maybe that step should be removed.

donjo: Ok, will edit the line in the release that says to install via the `npm install --save uswds@{{ version }}` command

maya:
For the future, let's try to use consistent branch names.. i.e. `release-1.4.4`.

donjo:
I propose we change the recommended action in that release step to reference what is being updated since using the word `release` alone implies that this project is receiving a new version when we're just updating a dependency. I don't know if the one I made up on the spot is the best option but I think that `release-1.4.4` would be unclear.

maya:
That works for me, maybe `update-uswds-v1.4.4`, `update-uswds-1.4.4`, or `update-v1.4.4`?

PR #472, opened 2018-01-03
Update analytics data

maya:
Adds 2017 Q2, Q3, and Q4 data; fixes 2017 Q1 data where new users and users appear to be reversed.

These reports add new USWDS websites to the query. 
  
[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/18f/web-design-standards-docs/update-analytics-data/whats-new/#web-analytics-for-sites-that-use-the-standards)

Note to team: @tdlowden suggested eliminating users and replacing with sessions if we have the choice: 
>"Users as a metric is pretty flawed and doesn't represent actual numbers of people. because it works with cookies, a single user visiting a site on a personal laptop, work laptop, personal phone, and tablet will appear as 4 distinct users." 

However, I suspect that the average person might not know what it is and a “user” is easy to wrap your head around. We can always add it in if we want to.


maya:
Should we add a note above the chart:
>**12% of all page views** to DAP-participating sites (~5,000) are coming via USWDS sites.

?

maya:
## Graph of data
chart

thisisdano:
Over a BILLION page views in 2017 Q2! 😮 

I don't think quite get the 
> 12% of all page views to DAP-participating sites (~5,000) are coming via USWDS sites.

Is this saying that USWDS-powered-site page views account for 12% of all DAP-collected gov page views — i.e. 12% of gov "traffic"?

Also, I'm inclined to leave the columns as is — perhaps total sessions could be a new column but I don't feel strongly about it. Whatever the "industry standard" measurement is, I'm for it.

maya:
>Is this saying that USWDS-powered-site page views account for 12% of all DAP-collected gov page views — i.e. 12% of gov "traffic"?

Yes, that's how I understood it.

OK to leave the columns as is, just wanted to float Tim's thoughts here.

PR #485, opened 2018-01-18
Strip vestigial ARIA and remove pseudostyling from labels in roadmap

thisisdano:
This removes unnecessary ARIA attributes from the roadmap list. It also removes hover states from labels, as there is no longer functionality that pseudostyling would support.

👀 [preview link](https://federalist-proxy.app.cloud.gov/preview/18f/web-design-standards-docs/dw-fix-roadmap-links/whats-new/product-roadmap/)

### DIFF
--- a/css/styleguide.scss
+++ b/css/styleguide.scss
@@ -909,10 +909,9 @@ h3 + .button_wrapper {
   background-color: $color-gray-warm-light;
 }
 
-// no such var
-//.usa-color-gray-cool-light {
-//  background-color: $color-gray-cool-light;
-//}
+.usa-color-gray-cool-light {
+  background-color: $color-gray-cool-light;
+}
 
 .usa-color-gold {
   background-color: $color-gold;
--- a/css/styleguide.scss
+++ b/css/styleguide.scss
@@ -418,38 +418,27 @@ body {
   position: absolute;
   text-decoration: none;
 
-  &:hover,
-  &:visited {
-    color: $color-gray-dark;
-  }
-
   &#{$external-href}::after {
     display: none;
   }
 }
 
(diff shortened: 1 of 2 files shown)


css/styleguide.scss
```diff
@@ -920,9 +909,10 @@ h3 + .button_wrapper {
   background-color: $color-gray-warm-light;
 }
 
-.usa-color-gray-cool-light {
-  background-color: $color-gray-cool-light;
-}
+// no such var
+//.usa-color-gray-cool-light {
+//  background-color: $color-gray-cool-light;
```
thisisdano: This line was killing my build because `$color-gray-cool-light` didn't seem to be available. I don't know why, so I commented out the whole rule. Brute force, but I wasn't sure this was an important rule for us.

maya: Isn't it used on the color palette page or did it change?

maya:
@thisisdano can you add a preview link to the PR?

thisisdano:
There should be one. Refresh maybe?

css/styleguide.scss
thisisdano: yeah, this broke that. I wonder why that var broke my build?

maya: Yeah needs to be added back in. See the screenshot from https://federalist-proxy.app.cloud.gov/preview/18f/web-design-standards-docs/dw-fix-roadmap-links/components/colors/#palette:
screen shot 2018-01-19 at 11 32 41 am

thisisdano: Found it. I modified that var by mistake in my utilities branch in the main repo.
Fixing

PR #488, opened 2018-01-22
Update jQuery to 3.3.1

maya:
Resolves a security vulnerability from an old jQuery dependency.


maya:
Running `npm update` will update `gulp-sourcemaps@1.12.1` (from 1.6.0). Lmk if we want to do that.

PR #498, opened 2018-02-15
Fix documented path to css and js

bennyty:
## Description

The [current documentation](https://designsystem.digital.gov/getting-started/developers/#installation) links to a likely outdated path: `css/lib/uswds.min.css` rather than the actual file at `css/uswds.min.css`. Similar for `js/uswds.min.js`.

### DIFF
--- a/pages/getting-started/developers.md
+++ b/pages/getting-started/developers.md
@@ -54,13 +54,13 @@ into your HTML pages:
 Add this to your `<head>` element:
 
 ```html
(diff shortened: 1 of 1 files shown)


bennyty:
I thought the same thing at first, however, the correct path would then be `lib/css/uswds.min.css`. I can see placing secondary lib folders in the css and js folders but for someone downloading and dropping it into a new project, imo this is better.

Thank you!

PR #505, opened 2018-02-22
Add NSF post

maya:
[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/add-nsf-post/whats-new/updates/2018/02/22/nsf-seed-fund/)

We can merge this now and publicize it (twitter) next week since we just posted an article two days ago.


donjo:
I noticed the top part of the interview refers to the USWDS by our new name but the interview content itself references the old name – do you think we should update the references or we keep it the same if the interview was conducted before the name change and want to keep it as close to reality as possible?

maya:
That's a good question. I don't know, happy to do it either way: change all references to "Design System" or put a disclaimer that the interview was done before the name change. Thoughts @thisisdano?

maya:
Changed to System and USWDS in 703f7f7.

PR #512, opened 2018-02-27
Update design assets

donjo:
Updates the assets from uswds/uswds-assets


maya:
Did we already run the release on `uswds-assets`? https://github.com/uswds/uswds-assets#generating-archive-files

donjo:
@maya yeah that was the command I ran to get the zipped assets that are included in this PR

maya:
So when you run the release on `uswds-assets` to generate the zip files, it adds the zip file to a different project `uswds-site` automatically? The process isn't super clear in the documentation.

donjo:
That was a manual part of the process that I had to do. I added a line in the asset readme to mention how the files make their way to this site in the [following PR](https://github.com/uswds/uswds-assets/pull/49). The release command just generates files for release but doesn't make any assumptions about how/where you will release (similar to the release command in the USWDS core project)

PR #520, opened 2018-03-17
Add themes to roadmap

### LINKED ISSUE #519: Add themes to product roadmap
Should we add themes back in the roadmap to more clearly convey the project's overarching goals? This was the previous format that was used and I think still applies to the work we're doing. Feels like adding these in connects it to user needs a bit more and helps you see the big picture. 

[:lock: Previous roadmap doc](https://docs.google.com/document/d/1HnDzyIqbEyQq7iN5Xv0rZ5ZrXdi-hlgtR0Rg-2veBZM/edit) - the only missing theme is from before was the "general public" or "using a federal website that incorporates the USWDS as a member of the public" which seems ok for now and can always be added in if needed.

screen shot 2018-03-14 at 4 52 33 pm

maya:
[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/add-roadmap-themes/whats-new/product-roadmap/)

Closes #519

### DIFF
--- a/_data/milestones.yml
+++ b/_data/milestones.yml
@@ -1,30 +1,39 @@
-- title: 2018 Product roadmap
-  id: 2018-roadmap
+- title: Deciding to use the design system
+  id: decide
   tasks:
     - title: Change our name to more accurately describe what we are and what we do
       status: Done
       url: https://github.com/uswds/uswds/issues/2300
-    - title: Develop an easier way to prototype and build consistently and incrementally
-      status: In progress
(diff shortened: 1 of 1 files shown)


maya:
Looking back at the old roadmap, seems like the themes were mentioned but not indicated on a per item basis. Instead, it was broken into 3 milestones ordered in how they'd be accomplished: https://web.archive.org/web/20161115123323/https://standards.usa.gov/about-our-work/product-roadmap/

Could be a better way of arranging it than this.

PR #529, opened 2018-03-27
Remove version from package.json

maya:
Doesn't seem like we need to version our site. Would also fix https://github.com/uswds/uswds/issues/2415.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "bugs": {
     "url": "https://github.com/uswds/uswds-site/issues"
   },
-  "license": "See LICENSE in LICENSE.md",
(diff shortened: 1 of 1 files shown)


hursey013:
I wonder if it makes sense to just remove `name`, `description`, and `version` and replace it all with `"private": true` so that it clear that it's not meant to be published (and will prevent it from ever accidentally happening).

maya:
@hursey013 sure, removed in 9aaaf00.

maya:
Good catch, added here: c311d54. And also sorted the list using [sort-package-json](https://www.npmjs.com/package/sort-package-json).

PR #533, opened 2018-03-28
Add name to Implementations page

maya:
Since we have multiple implementations of the same distribution methods (e.g. multiple WordPress themes) we need a way to identify them and ensure our sidebar navigation works.


maya:
Got a warning when I tried it so leaving off for now.

```
Liquid Warning: Liquid syntax error (line 27): Expected end_of_string but found pipe in "impl in site.data.implementations | sort: 'distribution'" in pages/getting-started/implementations.md
```

PR #536, opened 2018-04-13
Remove unused gems and plugin

maya:
We don't use redcarpet, rouge, or this redcarpet extender plugin, so this removes them.


Gemfile
```diff
@@ -2,12 +2,10 @@ source 'https://rubygems.org'
...
 gem 'json'
-gem 'redcarpet'
 gem 'open-uri-cached'
 gem 'jekyll-redirect-from'
 gem 'octopress-autoprefixer'
 
-gem 'rouge', '1.9'
 gem 'scss_lint', require: false
 
 gem 'jekyll-compose', group: [:jekyll_plugins]
```
hursey013: @maya Do you use `jekyll-compose` for anything?

maya: not sure what it's for or if I'm using it.

hursey013: Looks like its a tool for writing new posts and associating different states with them (published, unpublished, draft, etc). Looks like we are just adding new posts with regular jekyll way, so it should be safe to remove if we are looking to remove anything that's unused.

maya: Do you know what the `, group: [:jekyll_plugins]` does on the end of it?

hursey013: That just automatically includes the plugin in jekyll, it saves you a step of having to also include the name of the plugin in the `_config.yml`.

Issue #538, opened 2018-04-24
Use uswds README for Developers > Installation doc


hursey013:
The README for `uswds` very closely mirrors the page of content on https://designsystem.digital.gov/getting-started/developers/.

I'd like to propose integrating the actual README from uswds/uswds into the developer installation instructions in the same fashion that we build the Release Notes page so that we can avoid having to update the instructions in two places.

PR #539, opened 2018-04-27
Import uswds readme for Developers page

hursey013:
Import README from https://github.com/uswds/uswds/ into the developers page (to remove duplication).

Fixes https://github.com/uswds/uswds-site/issues/538

Preview: https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/bh-install-readme/getting-started/developers/


maya:
@hursey013 yes, let's add them to the README too. May have to combine the two JS sections into one.

hursey013:
@maya Updated the main repo's README to add the missing content: https://github.com/uswds/uswds/pull/2462

maya:
@hursey013 this still needs to address #1: The numeric steps in "Install using npm" all start at 1 instead of being an ordered list. Need to follow the formatting of the removed content to get it to display correctly.

Were you planning on opening a separate PR for it?

hursey013:
This just imports the README as is from `uswds`, so it is already a properly markdown formatted file, I don't have the ability to change the formatting of particular sections. I'm going to try tweaking the source README in `uswds` in order to add some additional indentation to the lists, and maybe that'll do the trick? Otherwise it's likely some difference between the way github and jekyll render markdown.

PR #542, opened 2018-05-04
Update dependencies

hursey013:
Update npm packages to latest versions.


hursey013:
Holding back the `axe-core` update since it seemed to cause some false positives, but from a quick check there appear to be no issues in updating the other dependencies.

PR #543, opened 2018-05-04
Remove politespace

hursey013:
We no longer use Politespace in our address form template, we only refer to it and link of to the vendor's website. This removes all of the politespace dependencies from our site.


hursey013:
@maya updated!

PR #551, opened 2018-05-25
Use uswds release-2.0 branch

hursey013:
Building off of the `add-utility-docs` this integrates the latest changes from the `release-2.0` uswds branch.

Since jekyll is no longer being used to process the sass, you need to run `npm run watch` in a separate terminal window if you are modifying any of the .scss files. In the future we can streamline this into the main `npm start` command if that's desirable.


hursey013:
I went ahead and also rearranged our dependencies in package.json to have only the code the winds up in compiled client side bundle to be in `dependencies` and everything else in `devDependencies`. Added an extra command to the `federalist` script to still install the `devDependencies` so our builds don't break. This will help avoid snyk flagging vulnerabilities that really shouldn't be considered risks.

PR #555, opened 2018-05-31
Add paragraph style docs

thisisdano:
[Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/dw-docs-paragraph/utilities/paragraph-style/)


maya:
The sidenav title says "Measure," but the actual title if "Maximum width (measure)." Should the sidenav be updated to match?

thisisdano:
Updated nav in b3fc04c

thisisdano:
@hursey013 I don't know why there are so many changes to the `package.lock` — I have no idea what that file does or what changes it, but I had a lot of shenanigans updating my node/npm and I want to make sure it didn't adversely affect something

hursey013:
@dano I think you should be fine, but if we run into any problems we can just remove the lock file and start fresh.

PR #556, opened 2018-05-31
[WIP] Add clearfix and display to utility docs

hursey013:
Still need to finish up `position` section of display page, struggling on how to demonstrate those best.

I also added a "Markup" section to some of the examples because without seeing the underlying code some of the examples wouldn't make any sense. Have we used this pattern in any of the other pages?

- [Clearfix](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/bh-clearfix-display/utilities/clearfix/)
- [Display](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/bh-clearfix-display/utilities/display/)


maya:
Some ideas for `position`, we could:
1. List the HTML as an MVP
2. Use an iframe to self-contain it
3. Have an image of what each thing does and link off to another page with a live example

maya:
We haven't used markup on any other pages. We sort of avoided it, by listing the class with the element in front of it here: https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/add-utility-docs/utilities/list-reset/

If we do use markup, I start to wonder if we should be pulling it in from fractal since it will generate the preview and markup by automatically and will be consistent with how it's done and looks on component pages.

hursey013:
@maya when we reviewed other utility documentation sites, one aspect that really appealed to me were the large code blocks that let you see what's going on in examples, or to quickly copy and paste code into your own project. I would definitely be in favor of showing the code, especially in situations without it there really isn't any other way to show an example. Take:
```
<div class="display-table">
  <div class="display-table-row">
    <div class="display-table-cell border-1px padding-2">.display-table-cell</div>
    <div class="display-table-cell border-1px padding-2">.display-table-cell</div>
  </div>
</div>
```
Visually, all you would see are the two `display-table-cell`'s, there would be no good way to demonstrate that `display-table` and `display-table-row` also are being used. Same goes for something like `clearfix`:
```
<div class="clearfix border-1px">
  <p class="float-left bg-base-lighter padding-2">.float-left</p>
  <p class="float-right bg-base-lighter padding-2">.float-right</p>
</div>
```
Pulling in the code from Fractal is a great idea if we are to go this direction.

maya:
@hursey013 FYI looks like there's a merge conflict that'll need to be cleared.

maya:
I'm getting the same error with `npm run axe && npm run crawl && bundle exec rspec` when I check out the commit before the merge conflict resolve.

hursey013:
I'll try sorting it out when I have a second, not sure what's going on.

PR #559, opened 2018-06-07
Update chrome remote interface

### LINKED ISSUE #550: Axe is not working locally
Error message below:
```
➜  web-design-standards-docs git:(add-utility-docs) npm test

> @ test /Users/mayaben-ari/Code/web-design-standards-docs
> npm run axe && npm run lint && npm run crawl && bundle exec rspec


> @ axe /Users/mayaben-ari/Code/web-design-standards-docs
> node config/run-axe.js

Static file server is listening at http://f09r9j-2jmg3qt:57877.
Chrome is debuggable on http://localhost:57878.
Running aXe on:
  /page-templates/docs/ A network request failed to load.
{ requestId: '51F129798E6A0124B92DFDA7559CC0E1',
  timestamp: 79243.403367,
  type: 'Document',
  errorText: 'net::ERR_NAME_NOT_RESOLVED',
  canceled: false }
Terminating with exit code 1.
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @ axe: `node config/run-axe.js`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the @ axe script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/mayaben-ari/.npm/_logs/2018-05-24T23_24_10_766Z-debug.log
npm ERR! Test failed.  See above for more details.
```

maya:
Resolves the issue outlined here: https://github.com/uswds/uswds/issues/2537

Fixes #550.

Axe is not working locally.

/cc @toolness

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,6 @@
     "chalk": "^2.4.1",
     "cheerio": "^1.0.0-rc.2",
     "chrome-launcher": "^0.10.2",
-    "chrome-remote-interface": "^0.25.5",
     "del": "^3.0.0",
     "express": "^4.16.2",
     "gulp": "^4.0.0",
@@ -98,5 +97,8 @@
       "build-js"
     ]
   },
-  "snyk": true
+  "snyk": true,
+  "devDependencies": {
+    "chrome-remote-interface": "^0.25.6"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "chalk": "^2.4.1",
     "cheerio": "^1.0.0-rc.2",
     "chrome-launcher": "^0.10.2",
+    "chrome-remote-interface": "^0.25.6",
     "del": "^3.0.0",
     "express": "^4.16.2",
     "gulp": "^4.0.0",
@@ -97,8 +98,5 @@
       "build-js"
     ]
   },
-  "snyk": true,
-  "devDependencies": {
-    "chrome-remote-interface": "^0.25.6"
-  }
+  "snyk": true
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -3365,14 +3365,12 @@
         "balanced-match": {
           "version": "1.0.0",
           "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-1.0.0.tgz",
-          "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c=",
-          "optional": true
+          "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c="
         },
         "brace-expansion": {
           "version": "1.1.11",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -98,5 +97,8 @@
       "build-js"
     ]
   },
-  "snyk": true
+  "snyk": true,
+  "devDependencies": {
```
thisisdano: any reason this is in dev dependencies and not in dependencies as the others? — think it may need to be in dependencies to make it work properly on federalist

maya: So this was just from running the same command as I did in the `uswds` repo: `npm i --save-dev chrome-remote-interface@latest`.

maya: In `uswds`, it was also saved to dev dependencies.

thisisdano: yeah, I _think_ this one needs to be different because of how the deps are loaded in federalist. I think it needs to be in regular deps and not dev deps

maya: It looks like Federalist built properly with no errors: https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/update-chrome-remote/ Is there another reason to change it besides this?

thisisdano: Not really — but we have all the others in dependencies. It is not a problem to have this one alone in devDependencies, but because it's different, it makes it seem as though it has some unique characteristic that the others do not. It makes you ask _why_ this one is here and the others are not. If there's a good reason to make it different, it should be different. If not, it should not.

maya: I've also asked @toolness and #federalist-support bc I'm not confident that I know the answer.

maya: @thisisdano Isn't the reason it's a dev dependency on the previous repo is to make sure Snyk doesn't scan it? We were getting a lot of non-relevant errors on `uswds` repo which is why Brian divided them into dependencies and dev dependencies to make sure only the real dependencies in production are scanned.

thisisdano: honestly, I think every dependency we have is a dev dependency, since all of it is in the service of testing, compilation, transpilation, etc, and none of it is required for the "app" to run

thisisdano: i had though the only reason we included stuff in dependencies was in some way related to the federalist build process — but that may not be the case

maya: @thisisdano we can also just make everything a  regular dependency now and make a note to switch everything out to a dev dependency later on. How does that sounds?

thisisdano: yeah — I _do_ think that everything should probably just be a devDependency (in here and in `uswds`, too), but it'll take a little bit of testing to make sure it works properly when we do that

thisisdano: I'm just gonna go ahead and pre-approve this.

maya: Updated in 61489b9.

PR #571, opened 2018-06-12
Add border utility docs

maya:
Adds border utility docs to site.

The title says "Border and outline" but there are no outline utilities. Should we just change this to "Border"? @thisisdano 

TODO: Add Default outputs to bottom of page

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/add-border-utility/utilities/border-and-outline/)


thisisdano:
There are outline utilities but it looks like I never got to them in the docs. `outline: tk` — yoops.
I'll add these to my list of things to add

maya:
I'm comfortable to have this be reviewed/merged now and add the outline in a separate PR if you are, since this already has a lot of work on it.

thisisdano:
I'm thinking I might change the base of this PR — and that of https://github.com/uswds/uswds-site/pull/556 — to my own docs branch `dw-componentize-docs` so I can merge and have all the work in one place before I PR into `add-utility-docs`

thisisdano:
The `border-(left|top|bottom|right)` examples are all currently showing `border-top`.

I have been arranging colors from lightest to darkest, including the base or core value (like `border-secondary` or `border-base`) in the center instead of at the start of the section. I find that gives a good sense of continuity without adding additional weight to the base value — which I don't necessarily think is more important than its variants.

The grayscale palette only includes
- white
- gray-5
- gray-10
- gray-30
- gray-50
- gray-70
- gray-90
- black

maya:
Fixed the order and removed unused gray in 4958ae3.

Currently, I only included border widths for one side (top). I expected people can take the formula at the top (and with the visual example of locations in #Border) and update it for their needs.

maya:
That said, I don't think it would add a huge amount to the page to add them in, I was trying to be cautious of the page being too long. We can add them if you find them useful.

maya:
Added all directional widths in 7456d58.

thisisdano:
I think I was unclear — I don't think we need to be explicit about the border widths on all sides. I think it worked OK as you had it originally. (It also works OK now, too, fwiw — you don't need to revert.)

What I'd meant to highlight was that in the first section, the borders aren't set properly. They are all showing `border-top` instead of what they claim to show:

screen shot 2018-06-12 at 12 08 58 pm

Also, now that I'm digging in, I realize there should be `border-x` and `border-y` utilities as well.

maya:
Ahhh got it, fixed in 8d14560 - thanks! 

Woops, yeah I guess I'll leave in the multi-directional borders, but also happy to remove it if it makes the page tighter.

thisisdano:
Thanks! Let's leave all the additional borders in for now — I've been erring on the side of more, and we can pare back if it becomes necessary

PR #582, opened 2018-06-27
Implement new colors and color values

thisisdano:
Dependent on https://github.com/uswds/uswds/pull/2566

- - -

Add new colors and values from USWDS
- Use correct hex values
- Assure all grade 50 colors are AA against white and black
- Sync theme colors with current USWDS defaults
- Add more darker and lighter variants for some theme families

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/dw-update-colors)

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
   "dependencies": {
     "jquery": "^3.3.1",
     "normalize.css": "^8.0.0",
-    "uswds": "github:uswds/uswds#dw-fix-colors"
+    "uswds": "github:uswds/uswds#release-2.0"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -49,7 +49,7 @@
   "dependencies": {
     "jquery": "^3.3.1",
     "normalize.css": "^8.0.0",
-    "uswds": "github:uswds/uswds#release-2.0"
+    "uswds": "github:uswds/uswds#dw-fix-colors"
```
maya: This needs to be changed back.

thisisdano: Yup, omw

PR #600, opened 2018-08-02
Add range slider

### LINKED ISSUE #595: Add range slider to docs site
https://github.com/uswds/uswds/issues/2374

maya:
On the next release, we'll be able to pull in the new range slider component. Hold merging until then, though the content can still be reviewed. Content used from [:lock: here](https://docs.google.com/document/d/1ymdyf1GxBxOwlv1BaaGfaxjHp5hKFu7RUCLum7WGMoQ/edit#heading=h.b5nqhvwoifvv).

Should the name of this be "Range input" or "Range slider"?

Resolves #595.

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/add-range-slider/components/form-controls/#range)

### DIFF
--- a/_components/form-controls/07-range-input.md
+++ b/_components/form-controls/07-range-input.md
@@ -7,8 +7,8 @@ lead: The range slider allows users to choose an approximate number from a range
 ---
 
 {% comment %}
-{% include code/preview.html component="range" %}
-{% include code/accordion.html component="range" %}
+{% include code/preview.html component="range-input" %}
+{% include code/accordion.html component="range-input" %}
 {% endcomment %}
 <div class="usa-accordion-bordered component-docs">
   <button class="usa-button-unstyled usa-accordion-button"
--- a/_components/form-controls/07-range-input.md
+++ b/_components/form-controls/07-range-input.md
@@ -1,5 +1,5 @@
 ---
-title: Range slider
+title: Range input
 parent: Form controls
 maturity: alpha
 order: 07
--- a/_components/form-controls/07-range-input.md
+++ b/_components/form-controls/07-range-input.md
@@ -29,7 +29,7 @@ Use a regular text input if a user needs to enter a precise number.
 ##### Guidance
 - The slider control should change color to indicate it is active when a user selects it.
 - Users should be able to drag the slider control or select somewhere along the slider itself to change the value.
-- Whenever appropriate, label each end of the slider with what it represents (whether a number range, “large” or “small,” “high” or “low”).
+- When appropriate, label the ends of the slider with the limits of the range (for example: "0/100", "small/large" or "less expensive/more expensive").
 - Set the `min` and `max` attribute of the `input` element to match whatever instructions or labels accompany the slider.
 - Set the `step` attribute so the slider is not too granular. Sliders should let users input approximate values. By setting the step to “10” or “20,” you keep users from inadvertently selecting values that are more precise than they intended.
 
--- a/_components/form-controls/07-range-input.md
+++ b/_components/form-controls/07-range-input.md
@@ -27,7 +27,7 @@ lead: The range slider allows users to choose an approximate number from a range
 Use a regular text input if a user needs to enter a precise number.
 
 ##### Guidance
-- When a user clicks the slider control to change their input, the slider control should change color to indicate it is active.
+- The slider control should change color to indicate it is active when a user selects it.
 - Users should be able to either click somewhere along the slider or drag the circular control to the left or right to change the  value. 
 - Whenever appropriate, label each end of the slider with what it represents (whether a number range, “large” or “small,” “high” or “low”).
 - Set the `min` and `max` attribute of the `input` element to match whatever instructions or labels accompany the slider.
--- a/_components/form-controls/07-range-input.md
+++ b/_components/form-controls/07-range-input.md
@@ -1,9 +1,10 @@
 ---
-title: Range input
+title: Range slider
 parent: Form controls
 maturity: alpha
 order: 07
-lead: The range input allows users to choose an approximate number from a range.
+id_custom: range
+lead: The range slider allows users to choose an approximate number from a range.
 ---
 
 {% include code/preview.html component="range-input" %}
--- a/_components/form-controls/07-range-input.md
+++ b/_components/form-controls/07-range-input.md
@@ -31,7 +31,7 @@ Use a regular text input if a user needs to enter a precise number.
 - Users should be able to drag the slider control or select somewhere along the slider itself to change the value.
 - When appropriate, label the ends of the slider with the limits of the range (for example: "0/100", "small/large" or "less expensive/more expensive").
 - Set the `min` and `max` attribute of the `input` element to match whatever instructions or labels accompany the slider.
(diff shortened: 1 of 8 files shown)


_components/form-controls/07-range-input.md
```diff
@@ -0,0 +1,40 @@
...
+    Documentation
+  </button>
+  <div id="range-docs" aria-hidden="false" class="usa-accordion-content" markdown="1">
+#### Accessibility
+- As you customize this form template, ensure it continues to follow the [accessibility guidelines for form templates]({{ site.baseurl }}/form-templates/) and the [accessibility guidelines for form controls]({{ site.baseurl }}/form-controls/).
+
+#### Usability
+##### When to use
+- When you expect a user will move the slider back and forth to experiment with different inputs.
+- When the exact number a user enters isn’t important, but you want to give them an opportunity to pick from a range.
```
thisisdano: This is more succinct. Is it clearer?

> - When the selection range is more important than the exact value the user selects.
> - When a relative value is more important than an exact value.

_components/form-controls/07-range-input.md
```diff
@@ -0,0 +1,40 @@
...
+
+##### When to consider something else 
+Use a regular text input if a user needs only to enter a precise number or only needs to enter a number once.
```
thisisdano: I don't really understand the second part here.

> Use a regular text input if a user needs ~~only~~ to enter a precise number. ~~or only needs to enter a number once.~~

_components/form-controls/07-range-input.md
```diff
@@ -0,0 +1,40 @@
...
+
+##### Guidance
+- When a user clicks the slider control to change their input, the slider control should change color to indicate it is active.
```
thisisdano: `The slider control should change color to indicate it is active when a user selects it.`

_components/form-controls/07-range-input.md
```diff
@@ -0,0 +1,40 @@
...
+- Users should be able to either click somewhere along the slider or drag the circular control to the left or right to change the  value.
```
thisisdano: For consistency: 

> Users should be able to drag the [slider] control ~~to the left or right~~ or select somewhere along the slider itself to change the value.

_components/form-controls/07-range-input.md
```diff
@@ -0,0 +1,40 @@
...
+- When you expect a user will move the slider back and forth to experiment with different inputs.
+- When the exact number a user enters isn’t important, but you want to give them an opportunity to pick from a range.
+
+##### When to consider something else 
+Use a regular text input if a user needs only to enter a precise number or only needs to enter a number once.
+
+##### Guidance
+- When a user clicks the slider control to change their input, the slider control should change color to indicate it is active.
+- Users should be able to either click somewhere along the slider or drag the circular control to the left or right to change the  value. 
+- Whenever appropriate, label each end of the slider with what it represents (whether a number range, “large” or “small,” “high” or “low”).
```
thisisdano: > When appropriate, label the ends of the slider with the limits of the range (for example: "0/100", "small/large" or "less expensive/more expensive").

_components/form-controls/07-range-input.md
```diff
@@ -0,0 +1,40 @@
...
+- Set the `min` and `max` attribute of the `input` element to match whatever instructions or labels accompany the slider.
```
thisisdano: This is a bit confusing, since `min` and `max` need to be numeric and we include nonnumeric examples. I don't know if we can realistically include guidance like this

_components/form-controls/07-range-input.md
```diff
@@ -0,0 +1,40 @@
...
+- Set the `step` attribute so the slider is not too granular. Sliders should let users input approximate values. By setting the step to “10” or “20,” you keep users from inadvertently selecting values that are more precise than they intended.
```
thisisdano: The guidance here only really works when the min is 0 and the max is 100. If the min/max were -1/1, this wouldn't help as much. Maybe it's more like `10% of the total range`?

> In a range slider, the relative value is more important than the specific value, so set the `step` attribute so it's not too granular. By setting the step to a value of 10-20% of the total range you prevent unnecessary precision and cognitive strain in your users.

_components/form-controls/07-range-input.md
```diff
@@ -0,0 +1,40 @@
+---
+title: Range input
```
thisisdano: I think I prefer `range slider` to `range input`, but I don't feel too strongly about it. But be consistent through out the page

_components/form-controls/07-range-input.md
```diff
@@ -0,0 +1,40 @@
...
+parent: Form controls
+maturity: alpha
+order: 07
+lead: The range slider allows users to choose an approximate number from a range.
```
thisisdano: `range slider` should be consistent with the component name? In this case, maybe `range input`?

_components/form-controls/07-range-input.md
maya: It was range slider originally but changed it to input bc of a duplicate id issue: the ID of the input was `range-slider` as was the ID of the section heading. We could change it to slider but would have to make a change to the ID on the USWDS repo for this component.

_components/form-controls/07-range-input.md
maya: We'll still need to explain to folks to set the `min` and `max` value to something in order for it to work. Can we rewrite this to: 

>Set the `min` and `max` attribute of the `input` element to correspond to the instructions or labels accompany the slider.

_components/form-controls/07-range-input.md
maya: Fixed in 9b4ee46.

_components/form-controls/07-range-input.md
maya: Fixed in 176357e36a158e865328325a387b7d51a6775706.

_components/form-controls/07-range-input.md
maya: Fixed in 718dd8d8407157746bca168f05e730a8637ca9d7.

_components/form-controls/07-range-input.md
maya: Fixed in a04c9972bc5652e4a23d5914cdba9b0d05373908.

_components/form-controls/07-range-input.md
maya: Updated in a42ccaf2c413889c5349f08d2e60202141531579.

_components/form-controls/07-range-input.md
maya: Updated in 57dd76050ada879e3103f7324f19c1bfb013d463.

_components/form-controls/07-range-input.md
maya: Updated in 0c9a544.

_components/form-controls/07-range-input.md
thisisdano: It seems like `range-slider` is the appropriate id for the component. Is there any way to use `Range slider` as the section name in the docs but give it a custom id?

maya: Added a custom control to change the section id in a9e0888 and updated the name to slider in 8546358.

PR #607, opened 2018-08-09
[WIP] Remove unused gem and update USWDS package

maya:


maya:
updating the USWDS package is kind of a moving target so going to close this to not get confused

PR #608, opened 2018-08-10
Set theme content styles to global in project settings

maya:
[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/update-settings/)


maya:
woohoo, this is working now after https://github.com/uswds/uswds/pull/2670 🎉

PR #612, opened 2018-08-10
Add spacing units section

maya:
Add spacing units page  to the Style section.

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/dw-add-style-section/style/)

Resolves uswds/uswds#2657


thisisdano:
@maya I scoped this PR to the spacing units page only.

maya:
Bc I opened the PR, GitHub won't let me review it. So this will be my review:

This looks good to me.

Some thoughts:
- Was a little unsure when you'd use the quoted format vs when you wouldn't. And they _why_ behind this.
- I don't know that we _need_ `margin-left:` on the utility functions and mixins. Maybe it would be easier to read without them?
- OR do we have the space to put the variable and/or function in the table above it? I'm imagining I'd use this page to pick the right space for my needs, and not have to do the mental translation to figure out which variable or function to use would be nice.

maya:
>I scoped this PR to the spacing units page only.

I think this also had type scale: https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/dw-add-style-section/style/type-scale/

PR #615, opened 2018-08-13
Fix bugs on v2 site

maya:
- Changes labels to tags
- Cleans up the layout docs page: de-nests markup and make it more robust to ensure h2's are the proper color style
- Ensures "Icons" page uses the correct page category so it has a sidebar nav appearing
- Refactor `css/_project-custom.scss`
  - reorganize code
  - fix several Sass errors 
  - fix ~50 stylelint issues (but not all) - 33 left


css/_project-custom.scss
```diff
@@ -1757,39 +1760,39 @@ html {
...
-    }
-  }
-  code { @extend .txt-code; }
-  aside {
-    code {
-      @include u-bg($gray-warm-10);
-    }
-  }
-}
+// .usa-body-content, .usa-prose {
```
maya: @thisisdano can you confirm this Sass code is not needed? It didn't seem to be in use and the `.usa-prose` here was adding confusion when I was trying to test the default behavior.

thisisdano: 👀

thisisdano: This is not necessary any longer — I think it was a first draft of mine of what became your work.

maya: Removed! is any of the CSS above this starting from `CUSTOM PROJECT SASS` in use too? Starting from https://github.com/uswds/uswds-site/pull/615/files#diff-8ef0ab0a17bcd64e66ebd7c22ee57320R1633 and up to this code.

thisisdano: I think `.is-inverse` is the only rule worth keeping in that section

PR #619, opened 2018-08-14
[ci] Fail CI on SCSS compile error

el-mapache:
## Description

Since we are running our gulp tasks in parallel, scss failing to compile won't cause the build step to fail on CI. It will fail subsequent steps when the css file isn't found. This can be confusing. So, fail the build immediately if scss cant compile


thisisdano:
@el-mapache It's nice, when running `npm run watch`, to keep the watch script running through the sass fail so we don't have to keep restarting it. But we _do_ want that hardfail in the federalist build. Is there any good way to apply this bailout only when running the federalist build script?

el-mapache:
@thisisdano Good point, I didn't check the watch. I can add a `development` env variable which bypasses the call to `process.exit`

el-mapache:
@thisisdano would you mind taking a peek and see if this does what you want? I tested it and it seems to keep the `watch` running even if the scss is bogus, which I believe is the desired behavior!

thisisdano:
This looks great from the `watch` side — I haven't tested on federalist, but if you have, this is good to go.

el-mapache:
@thisisdano I havent tested it there, but nothing changed from that perspective beyond the change to fail the build if the scss is invalid.

PR #620, opened 2018-08-14
Fix remaining stylelint issues

maya:
Fixes additional stylelint warnings in `css/_project-custom.scss`

The issues left in `css/_project-custom.scss` are related to ColorKeyword and ColorVariable that were caused by this: `@include u-text(white);`
```
[19:41:05] _project-custom.scss:2041 [W] ColorKeyword: Color `white` should be written in hexadecimal form as `#ffffff`
[19:41:05] _project-custom.scss:2041 [W] ColorVariable: Color literals like `white` should only be used in variable declarations; they should be referred to via variable everywhere else.
```

What should we do about these @thisisdano? We probably have a lot of these in the main repo, or will.
 
---
In other files, there are more stylelint warnings, which tells me we're not linting these files in the uswds main repo bc they were copied from it:
```
Comment: Use `//` comments everywhere
[20:01:00] 18 issues found in /Users/mayaben-ari/Code/web-design-standards-docs/css/_uswds-project-settings.scss
[20:01:00] _uswds-project-settings.scss:1 [W] Comment: Use `//` comments everywhere
```

```
SpaceBetweenParens: Expected 0 spaces after `(` instead of ` `
[20:01:00] 100 issues found in /Users/mayaben-ari/Code/web-design-standards-docs/css/_uswds-project-utilities-settings.scss
[20:01:00] _uswds-project-utilities-settings.scss:558 [W] SpaceBetweenParens: Expected 0 spaces after `(` instead of ` `
```

### DIFF
--- a/.scss-lint.yml
+++ b/.scss-lint.yml
@@ -14,7 +14,7 @@ linters:
   ColorVariable:
     enabled: true
   Comment:
-    enabled: true
+    enabled: false
   DebugStatement:
     enabled: true
   DeclarationOrder:
--- a/css/_project-custom.scss
+++ b/css/_project-custom.scss
@@ -400,15 +400,13 @@ h2 + .tooltip {
   margin-top: 0;
 }
 
-a {
-  &.cta {
-    img {
-      height: 1.5rem;
-      margin-right: 0.4rem;
-      position: relative;
-      top: 2px;
-      vertical-align: baseline;
-    }
+.cta {
+  img {
+    height: 1.5rem;
+    margin-right: 0.4rem;
+    position: relative;
+    top: 2px;
+    vertical-align: baseline;
   }
 }
 
@@ -1042,14 +1040,8 @@ code {
     letter-spacing: -0.2em;
     content: '\00a0';
   }
-}
 
-.highlight {
-  max-width: $lead-max-width;
-}
-
-a {
-  code {
+  a & {
     @include u-border(1px, $blue-50v);
     @include u-bg(transparent);
 
@@ -1058,6 +1050,10 @@ a {
       @include u-text(white);
     }
   }
+}
+
+.highlight {
+  max-width: $lead-max-width;
 }
 
 // Code snippet boxes ----------------- //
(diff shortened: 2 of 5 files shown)


thisisdano:
In the short term, at least, the `white` issue could be resolved by writing it as a string, as in `color('white')` — and this is probably a good practice for all the strings we're using in functions and mixins anyway.

We should be linting all the files the main repo — I wonder why some of these seem to be skipped. 

I'm not sure I agree with the `Use `//` comments everywhere` rule, either. It's not a big deal, but it feels like that's a bit too strict.

maya:
>We should be linting all the files the main repo — I wonder why some of these seem to be skipped.

It's bc they were never added to the linter, those files would need to be added to what the linter watches.

maya:
After adding the quotes to `('white')` in 40ad55b like you mentioned, I'm getting this Sass error:
```
Message:
    node_modules/uswds/dist/scss/core/mixins/utilities/_text.scss
Error: `white` is not a valid `text` value.
        on line 60 of node_modules/uswds/dist/scss/core/mixins/utilities/_text.scss, in mixin `u-text`
        from line 1053 of css/_project-custom.scss
        from line 8 of css/styleguide.scss
>>       @error "`#{$this-value}` is not a valid `text` value.";

   ------^
```
@thisisdano could you advise?

thisisdano:
OK — this means I'll need to go in and check the values in the tokens.

thisisdano:
Yeah, I have to look at this more carefully, from what I can tell, `'white'` should work as a color value, so I have to see what's going wrong

maya:
@thisisdano thanks, this is working now 👍 

Also, I've disabled the Comments lint rule in 50bc994 since they are stripped out.

There are officially 0️⃣ scss lint errors now 🎉

maya:
I would love if we had a notification for lint issues, like on our other repo, that would pass CircleCI if there are issues, bc otherwise there's no way to see this and errors will be merged until we manually look at this again.

PR #621, opened 2018-08-16
Add type scale docs

thisisdano:
Resolves https://github.com/uswds/uswds/issues/2658

Preview TK


maya:
@thisisdano Should we add a heading to the utility class column in the table? For example:

screen shot 2018-08-16 at 12 27 50 pm

thisisdano:
@maya Added that heading as `example utility class` just to clarify that that's not the only utility class for that scale.

PR #633, opened 2018-08-27
Contrast checker POC

el-mapache:
## Basic contrast checker utility

## Description

Really simple POC contrast checker. Currently, all colors in all families are checked against one another, assuming they have a magic number difference of > 30.

I'm not checking the colors against white or black, which is probably something we'll need to add as well?

### Instructions:
* If pulling down for the first time, be sure to run `npm install`!

#### See all failing contrasts between grades in all color families
* `npm run check-contrast -- --c` 

#### See all failing contrasts between all grades in a single family, plus white and black
* `npm run check-contrast -- --c --f ${family-name}`

#### Apply gray luminance values to all matching grades in another color family
* `npm run check-contrast -- --al --f ${family-name}`

#### View list of relative luminance values for all colors in a given family
* `npm run check-contrast -- --l --f ${family-name}`

#### To add optional debug output
* use the `--d` flag in the above commands


maya:
The original check also still works (all color families checked against each other):
- `npm run check-contrast`
- Open `contrast-report.json`

el-mapache:
@maya I actually commented out that code, so the file you are seeing is probably the old one from a previous run

maya:
Oh I see, is there any harm to have it back in there?

el-mapache:
@maya its not using the same algorithm as the latest code, so I think there are might be some false positives in the output. I want to get the code we have now cleaned up and then go back and make the original check use that.

PR #642, opened 2018-09-04
Update warning variable name

maya:
Documentation changes that go with the changes in https://github.com/uswds/uswds/pull/2714.


maya:
Done in 7214850.

PR #644, opened 2018-09-10
Add new typography settings

thisisdano:
This will make the site look normal once https://github.com/uswds/uswds/pull/2731 goes through


maya:
@thisisdano this still has WIP in the title. Is it ready for review?

thisisdano:
Sorry, it should be all good now

PR #647, opened 2018-09-17
Update old grid to new grid

maya:
...with the exception of:

- components/grid - redirects to the v2 grid
- components/color - redirects to v2 utilities/color

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/update-grid-v2)


_data/nav.yml
```diff
@@ -23,9 +23,11 @@ Components:
   - text: Overview
     href: /components/
   - href: /components/typography/
-  - href: /components/colors/
+  - text: Colors
+    href: /components/colors/
   - href: /components/icons/
-  - href: /components/grids/
+  - text: Grids
+    href: /components/grids/
```
maya: Another option is we could also change this to the `utilities/*` links so it's not a redirect and goes directly. Thoughts?

_includes/banner.html
```diff
@@ -2,7 +2,7 @@
 <section class="usa-banner site-banner">
```
maya: Can eventually use the v2 banner + `site-banner` class.

_data/nav.yml
thisisdano: I might suggest moving colors out of components altogether

thisisdano:
I don't think I'm seeing the redirects working.

maya:
Ah, OK. They were working before, it might be bc I force added the titles in 08ce2d3.

_data/nav.yml
maya: Removed "Colors" in c750d57.

maya: Keep grids in components?

maya:
@thisisdano redirects should be working now. Take a look now.

PR #654, opened 2018-09-26
Remove bourbon and update some v1 styles

thisisdano:
🔎 [Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/dw-remove-bourbon/)

- - -

This PR removes Bourbon from the site and updates all the custom code to v2 tokens and palette. There are a couple quirks, mostly around collapsing columns and header top margin, but this a bigger issue for another day.

A few items will shift around (the hero is probably the biggest difference) — but I'm actually a bit surprised just how similar everything looks. I've been able to activate `$respect-user-font-size` as well, which is nice.


maya:
Are these style changes intentional?

After
screen shot 2018-09-26 at 2 25 11 pm


Before 
screen shot 2018-09-26 at 2 25 48 pm

maya:
Is the increase in the lead text size intentional:

After:
screen shot 2018-09-26 at 2 28 39 pm

Before:
screen shot 2018-09-26 at 2 28 58 pm

maya:
Gap in nav current/hover state:
screen shot 2018-09-26 at 4 06 01 pm

thisisdano:
Yeah, I'm trying to address that gap, but it's going to take a sec — I have to import the technique from the standard header

maya:
Which style of the card looks the best:

## Current
screen shot 2018-09-26 at 4 28 01 pm

## Current (without added margins)
screen shot 2018-09-26 at 4 27 41 pm

## Wider (similar to the existing site)
screen shot 2018-09-26 at 4 27 09 pm

thisisdano:
I'd say option 2 — I can implement that.

maya:
Since that's the default hero style, we just have to remove our custom CSS.

thisisdano:
OK, done.

css/_project-custom.scss
```diff
@@ -1862,3 +1836,24 @@ li {
...
   }
 }
+
+p + .usa-accordion,
+h2 + *,
+h3 + *,
+h4 + *,
+h5 + *,
+h6 + * {
+  margin-top: units(2);
```
maya: This code seems to impact the top heading margins (1.5em) when another heading element follows it.  See "Spacing" https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/dw-remove-bourbon/components/typography/#typesetting

css/_project-custom.scss
```diff
@@ -1862,3 +1836,24 @@ li {
...
+}
+
+.usa-sidenav a {
+  padding-right: units(0.5);
```
maya: What is this code used for?

thisisdano: For reducing the right padding on the nav items, since it felt like there was too much — it was causing some line breaks that didn't feel strictly necessary.

maya: For the site sidenav or mobile nav or both? Could you scope it to the site elements and not the sidenav preview?

thisisdano: What's the best way to scope to site here?

thisisdano: Wonder if there's a `:not()` selector solution here that could also help us think about our prose scope bleed into component previews.

thisisdano:
@uswds/core 
I'd like to suggest that we hold this PR until after the beta release — then we can connect the `package.json` ref to the beta release, pass circle, and do a realistic test for using the beta.

maya:
It's interesting that this shows a Snyk vulnerability introduced in the uswds repo via mem@1.1.0, may want to take a look at the uswds repo and use at least mem@4.0.0.

PR #695, opened 2018-12-17
Fix flex-justify example and section closing tag

blacktm:


thisisdano:
Good catch — we'll work to get that in asap

PR #714, opened 2019-03-08
Use white background for demo pages

thisisdano:
[Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/dw-scope-bg-color/page-templates/docs/)

- - -

Uses a white background for all `body.layout-demo` pages.

- - -

Fixes #714

### DIFF
--- a/css/_uswds-theme-custom-styles.scss
+++ b/css/_uswds-theme-custom-styles.scss
@@ -80,6 +80,7 @@ $site-background-color: 'gray-2';
 
 body {
   @include u-bg($site-background-color);
+
   &.layout-demo {
     @include u-bg('white');
   }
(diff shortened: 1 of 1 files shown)


css/_uswds-theme-custom-styles.scss
```diff
@@ -80,6 +80,10 @@ $site-background-color: 'gray-2';
 
 body {
   @include u-bg($site-background-color);
+
+  &.layout-demo {
+    @include u-bg('white');
+  }
```
maya: How about just adding the gray on `.default-container` and remove the override?

thisisdano: OK, done

Issue #745, opened 2019-03-28
Typography size inconsistencies


maya:
Using the web inspector to view the font size, many of them did not match up with the actual values on the docs site and fractal site with the written values.

<img width="1248" alt="Screen Shot 2019-03-28 at 12 30 31 PM" src="https://user-images.githubusercontent.com/5249443/55187120-6def9a80-5155-11e9-8571-ca674e7ec23d.png">

thisisdano:
Hm. Let's see what's up here...

thisisdano:
Ugh, I think this another instance of "site defaults do not match system defaults`.

The system theme font size defaults are:
```
$theme-type-scale-3xs:      2;
$theme-type-scale-2xs:      3;
$theme-type-scale-xs:       4;
$theme-type-scale-sm:       5;
$theme-type-scale-md:       6;
$theme-type-scale-lg:       9;
$theme-type-scale-xl:       12;
$theme-type-scale-2xl:      14;
$theme-type-scale-3xl:      15;
```

but the site defaults are
```
$theme-type-scale-3xs:      2;
$theme-type-scale-2xs:      3;
$theme-type-scale-xs:       4;
$theme-type-scale-sm:       5;
$theme-type-scale-md:       6;
$theme-type-scale-lg:       8; // different from system default
$theme-type-scale-xl:       11; // different from system default
$theme-type-scale-2xl:      14;
$theme-type-scale-3xl:      15;
```

Additionally, the site data lists the default tokens as something different from wither of these. So, first, I'll update the site data to match the system defaults — then work out something where we can get all of them in sync.

maya:
Some size issues on the fractal site which shouldn't be affected by theme stuff: https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/release-2.0/components/preview/typesetting.html

h3 (`font-serif-lg`) listed as 23.4px, actually is 21.4px
h5 listed as 14.6px, no near equivalent in docs
h6 (`font-serif-2xs`) is listed as 13.7px, actually is 13.8px 

The rest seem accurate.

thisisdano:
Once the docs change is in the `h` values should be in sync:
h1: 2xl
h2: xl
h3: lg
h4: sm
h5: xs

maya:
What are h6 and body font (p)? 2xs and xs?

thisisdano:
`p` is probably set to `sm` (16px) — but I think in most cases it takes the normalized equivalent of 1rem.

From typography settings:
[...]

PR #746, opened 2019-03-28
Fix font-size output in docs

thisisdano:
[Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/dw-font-size/utilities/font-size-and-family/) ✨ 

This should update the docs to use the proper USWDS defaults and output the proper utility into the example to display the proper size.

Fixes uswds/uswds-site#745.

### DIFF
--- a/_components/utilities/font-size-and-family.md
+++ b/_components/utilities/font-size-and-family.md
@@ -135,7 +135,7 @@ utilities:
               {{ output }}px
             {% endcapture %}
             {% capture this_example %}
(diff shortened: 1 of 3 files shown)


maya:
Type-based size and family utilities are correct.

Role-based size and family utilities are not showing up.

thisisdano:
Aw shoot right. Thats because we don't output role-base system token utilities.... hm. 🤔

thisisdano:
Ugh, this is annoying.

thisisdano:
OK — I think I've got it now.

PR #752, opened 2019-04-04
Use 2.0.0 and update refs

thisisdano:
Hold until we update the uswds package


maya:
Is there a reason not to use `{{ site.data.uswds_version }}` so we don't have to do this manually?

thisisdano:
Yes, we should use `{{ site.data.uswds_version }}` now

PR #755, opened 2019-04-05
Add search to the site

maya:
Uses https://github.com/18F/jekyll_pages_api_search.

Note: All of our redirects are being crawled and used in the results so you'll see some empty spots in the results when we have redirects. Wondering if these redirects can be removed as I'm not sure how used they are.

## Regular page
Screen Shot 2019-04-11 at 12 45 13 PM

## Search results page
Screen Shot 2019-04-11 at 12 45 35 PM

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/add-search/)


thisisdano:
:tada:

PR #768, opened 2019-04-10
Update "form controls" side nav to match headings

blacktm:
This updates the [form controls](https://v2.designsystem.digital.gov/components/form-controls/) side nav to match the page headings, namely "Text inputs and area" 👉 "Text input" and "Checkboxes" 👉 "Checkbox" (fixing broken anchor link).

PR #769, opened 2019-04-10
Fix form sidebar links

maya:
Implements fixes in #768 but on the v2 branch.

cc @blacktm


maya:
Since I'd already reviewed @blacktm's work in #768, think this is OK to self-review.

PR #780, opened 2019-05-31
v1: Update banner URL

thisisdano:
Update the v2 banner link to `designsystem.digital.gov` from `v2.designsystem.digital.gov`


jeremyzilar:
Looks great 🎉

PR #838, opened 2019-12-02
Improve layout code

thisisdano:
[Preview](https://cg-06ab120d-836f-49a2-bc22-9dfb1585c3c6.app.cloud.gov/preview/uswds/uswds-site/dw-display-block/) ✨ 

OK, let's do this a little better this time. 
- [x] Preview the changes in Federalist
- [x] Test navigation and display in Chrome and FF

This change does a couple things:
- Removes the `inline-block` declaration from `main-content`
- Adds padding instead of margin to `main-content`
- Removes the float from `sidenav`
- Adds a `grid-row` to the `default-container` to put it into flex mode


_layouts/post.html
```diff
@@ -2,10 +2,10 @@
 layout: default
 category: What’s new
 ---
-<div class="default-container">
+
+<div class=" grid-row">
```
mejiaj: Nitpick, but extra space here.

_layouts/post.html
```diff
@@ -14,15 +14,17 @@
 
   <main id="main-content" class="main-content">
```
mejiaj: I know we're not using `position: absolute;` anymore for the component page, which was having that `display: inline-block` issue where the menu wasn't clickable on firefox.

It could be the time to reevaluate if we need positioning on it at all. I say this because if you inspect `.main` on the pages w/ 2 columns you see this:

image

The code:
```css
//
// _uswds-theme-custom-styles.scss:470
//
.main-content {
  @include u-pin("right");
  @include u-pin("bottom");
  padding-top: units(5); // starting: not scrolled
  position: relative;
  width: 100%;

  @include at-media("desktop") {
    width: calc(100% - #{units($width-nav-sidebar)});
  }

  .lt-ie9 & {
    width: 75%;
  }
}
```

thisisdano:
Unfortunately, this work still does not address the [Firefox printing bug](https://bugzilla.mozilla.org/show_bug.cgi?id=939897). According to the bug page this is now a high priority bug for Firefox. More info here: https://wiki.mozilla.org/Platform/Layout/Printing_and_fragmentation

Until this is addressed, we will see issues with printing any content contained in a flex container.

PR #856, opened 2020-01-28
use direct link to editing showcase page

afeld:
I think folks will have a hard time finding the page to edit otherwise.

### DIFF
--- a/pages/documentation/getting-started/showcase-all.md
+++ b/pages/documentation/getting-started/showcase-all.md
@@ -3,7 +3,7 @@ permalink: /getting-started/showcase/all/
 layout: styleguide
 title: USWDS powered sites
 category: How to use USWDS
-lead: Below are a list of website and applications currently using USWDS. If your project is currently using USWDS and you do not see it on this list, please feel free to [submit a pull request](https://github.com/uswds/uswds-site/blob/master/pages/documentation/getting-started/showcase-all.md) or email the core team at [uswds@support.digitalgov.gov](mailto:uswds@support.digitalgov.gov).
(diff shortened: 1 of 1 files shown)


pages/documentation/getting-started/showcase-all.md
```diff
@@ -3,7 +3,7 @@ permalink: /getting-started/showcase/all/
 layout: styleguide
 title: USWDS powered sites
 category: How to use USWDS
-lead: Below are a list of website and applications currently using USWDS. If your project is currently using USWDS and you do not see it on this list, please feel free to [submit a pull request](https://github.com/uswds/uswds/pulls/) or email the core team at [uswds@support.digitalgov.gov](mailto:uswds@support.digitalgov.gov).
+lead: Below are a list of website and applications currently using USWDS. If your project is currently using USWDS and you do not see it on this list, please feel free to [submit a pull request](https://github.com/uswds/uswds-site/blob/master/pages/documentation/getting-started/showcase-all.md) or email the core team at [uswds@support.digitalgov.gov](mailto:uswds@support.digitalgov.gov).
```
thisisdano: This is an autogenerated page from this list: https://github.com/uswds/uswds/blob/develop/docs/WHO_IS_USING_USWDS.md

Even harder to find! Let's link to that page instead.

afeld: Updated! That said, not sure it's worth the complexity and user confusion of having the list in two places…you may consider consolidating.

thisisdano: Agreed! Having this list in this repo is actually really unhelpful, in addition to being confusing, as it connects content to source code for content that has no _intrinsic_ relationship. This is a good prompt to fix it and move the showcase to `uswds-site`

thisisdano: Added as issue: https://github.com/uswds/uswds/issues/3324

PR #892, opened 2020-03-24
USWDS-3348: create button groups page

jaredcunha:
[Preview](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.app.cloud.gov/preview/uswds/uswds-site/10x-forms/button-groups/components/button-groups/) → 

## Description

Details can be found in https://github.com/uswds/uswds/issues/3348

## Additional information

Screenshot_2020-03-26 Button groups United States Web Design System

- [ ] Run `npm test` and make sure the tests for the files you have changed have passed.


jaredcunha:
Documentation is ready for some eyes. Have to find where tests are failing.

jaredcunha:
@thisisdano Made a pretty significant change here to distinguish guidance between the two button group variants. I'm still waiting for more feedback on how to handle how we'd stretch to the container. I kinda liked [Maya's suggestion](https://gsa-tts.slack.com/archives/C3F14AHSQ/p1586800298069700?thread_ts=1586800012.069000&cid=C3F14AHSQ). 

I think in the set of segmented controls, I might just show a single example - the outline buttons with a single default button as the current state, as opposed to what appears in the screen shot.

Screenshot_2020-04-13 Button groups United States Web Design System

_components/button-groups/01-default.md
```diff
@@ -0,0 +1,55 @@
+---
+title: Default button group
```
Ben-Judy-Flexion, line 2: Total nit-pick, but I prefer consistency in the titles. This one is "Default button group" while the other is simply, "Segmented."
I suggest either including or excluding "button group" in both titles for the subsections.
My preference is to match the titles exactly to the nav, so they would be "Default" and "Segmented."

_components/button-groups/01-default.md
```diff
@@ -0,0 +1,55 @@
...
+    <div id="accordion-bordered-docs" aria-hidden="false" class="usa-accordion__content site-component-usage">
+      <h4>When to use the default button groups</h4>
+      <ul class="usa-content-list">
+        <li><strong>Actions have a contextual relationship.</strong> For example, the default button group can be used when a form has both a primary and alternative action.</li>
+      </ul>
+      <h4>When to consider something else</h4>
+      <ul class="usa-content-list">
+        <li><strong>If actions are not related.</strong> Consider how placement and alternative structure of unrelated actions can improve usability over placing all actions in a group.</li>
+        <li><strong>When mixing destructive and non-destructive actions.</strong> This can lead to making mistakes.</li>
+        <li><strong>Navigational links.</strong> Though buttons can be used for links, buttons in a group should relate to actions.</li>
```
Ben-Judy-Flexion, line 26: So we are recommending against using button groups for "Back / Next" navigation? What if that is the primary action -- for users to step through a linear set of content, such as a tutorial or gallery?

_components/button-groups/01-default.md
jaredcunha: Yeah, I went back and forth on that yesterday but I think I agree with you

_components/button-groups/01-default.md
jaredcunha: @Ben-Judy-Flexion I think that's an appropriate use case… but wasn't sure how best to differentiate between the back/next, save&continue/skip, etc variety of links vs something like "view article." Do you have any suggestions?

Ben-Judy-Flexion: Hmm. What about:

**Linking to content.** Regular links, not buttons, should be used for navigating through site pages. However, buttons in a group can be used for directional navigation and actions (e.g., "Back," "Next," "Continue," "Skip," "Cancel," "Share.")

jaredcunha: I think I like that phase, "linking to content." I'd just modulate a little by saying buttons in button groups instead of all buttons. Totally normal for buttons to be used in directing users to a page if that is a primary called to action (USWDS even has examples of this).

Ben-Judy-Flexion: Is updating the existing Button guidance on the table? Because this is confusing to me, then.

image

thisisdano: Existing guidance is on the table

jaredcunha: For clarity, I was referring to these instances
Screen Shot 2020-04-14 at 4 31 35 PM

_components/button-groups/01-default.md
jaredcunha: @Ben-Judy-Flexion This is resolved now

_components/button-groups/01-default.md
jaredcunha: I've updated the guidance here.

_components/button-groups/01-default.md
```diff
@@ -0,0 +1,56 @@
...
+      <h4>When to use the default button groups</h4>
+      <ul class="usa-content-list">
+        <li><strong>Actions have a contextual relationship.</strong> For example, the default button group can be used when a form has both a primary and alternative action.</li>
+        <li><strong>Stepping through linear content.</strong> Buttons in a button group can be used for directional navigation and actions (e.g., "Back," "Next," "Continue," "Skip," "Cancel.").</li>
+      </ul>
+      <h4>When to consider something else</h4>
+      <ul class="usa-content-list">
+        <li><strong>If actions are not related.</strong> Consider how placement and alternative structure of unrelated actions can improve usability over placing all actions in a group.</li>
+        <li><strong>When mixing destructive and non-destructive actions.</strong> This can lead to input mistakes.</li>
+        <li><strong>Linking to content.</strong> Buttons in buttons button groups should not be used for navigating through site pages.</li>
```
thisisdano, line 27: @Ben-Judy-Flexion Can you clarify this one? Does this mean to discourage uses like `Next page` and `Previous page` button links?

_components/button-groups/02-segmented.md
```diff
@@ -0,0 +1,53 @@
...
+        <li><strong>If there are more than three buttons.</strong> Be mindful of how a long list of buttons might appear on small screens. An alternative type of control might be more suitable.</li>
+      </ul>
+      <h4>Usability guidance</h4>
+      <ul class="usa-content-list">
+        <li><strong>Avoid ambiguity of current state.</strong> Make sure current states are clearly communicated and understood.</li>
+      </ul>
+      <h4 class="usa-heading">Accessibility</h4>
+      <ul class="usa-content-list">
+        <li><strong>Use <code>aria-label</code> to give the buttons a useful name.</strong> Some contexts may require additional context provided to screen readers.</li>
+        <li><strong>Use the <code>&lt;button&gt;</code> element.</strong> Do not use <code>&lt;a&gt;</code> or <code>&lt;span&gt;</code> because screen readers will not know if it is a usable button.</li>
```
thisisdano, line 34: We allow `<a>` elements to be styled as `usa-button`s. Is this a mistake?

_components/button-groups/02-segmented.md
```diff
@@ -0,0 +1,53 @@
...
+      </ul>
+      <h4 class="usa-heading">Implementation</h4>
+      <ul class="usa-content-list">
+        <li>Do not mix differently-sized buttons in the same group.</li>
```
thisisdano, line 38: Does this mean something like mixing `usa-button` and `usa-button--big`?

_components/button-groups/02-segmented.md
```diff
@@ -0,0 +1,53 @@
...
+        <li>Use short, descriptive button labels to describe controls.</li>
+        <li>Use default buttons as a current state and outline buttons as a non-current state. The segmented button group works with all button variants. However, you will need to create custom styles to distinguish state for all other button variants.</li>
```
thisisdano, line 40: @jaredcunha Does `you will need to create custom styles to distinguish state for all other button variants` mean that you will need custom styles for all _non-primary_ variants?

_components/button-groups/01-default.md
Ben-Judy-Flexion: I think our intent is to advise against using buttons when regular ol' text links would do better. Next and Previous are okay when content is organized sequentially. We iterated on the wording of this but it might need some more clarification. Maybe:

> **Linking to content.** Buttons in buttons button groups should not be used for navigating to site pages when a text link would contextually provide a better indication of the content. Grouped buttons such as "Next" and "Previous" are okay when content is organized sequentially.

...or something like that?

_components/button-groups/02-segmented.md
jaredcunha: @Ben-Judy-Flexion Yes - this more about the button component than the button group component. There are no outline styles available for the other types of buttons. So currently, if you want to use button group with the cool accent blue button, you have to create an outline version of that button because one does not exist.

_components/button-groups/02-segmented.md
jaredcunha: Yes. I guess same would be for mixing `usa-button` and `usa-button--small` if that gets added.

_components/button-groups/02-segmented.md
jaredcunha: @thisisdano I should reword as something along the lines of "do not use `a` because it is a link. Do not use `span` because screen readers will not know if it is a usable button."  In the case of the segmented group, I think this is more about the semantic context for the button group vs buttons in general.

_components/button-groups/02-segmented.md
thisisdano: Would this be better or more clear? `However, you may need to create custom styles to distinguish state for some button variants.`

_components/button-groups/01-default.md
jaredcunha: @thisisdano I think what @Ben-Judy-Flexion mentioned is what we're trying to get at.

_components/button-groups/02-segmented.md
jaredcunha: @thisisdano I think that works.

_components/button-groups/02-segmented.md
thisisdano: If we allow something like `Next/Previous` might it make sense to use `a` in that instance? I'm comfortable with encouraging buttons and disallowing `span`s but I'm inclined to cut the part about links

thisisdano: Ah but this is for segmented. I'm fine with it as is.

thisisdano: That is, I'm fine with the edit you proposed @jaredcunha

thisisdano:
I'll make these copy edits

PR #906, opened 2020-03-31
Update site name and description

afeijoo:
USER STORY
As the design system, I want external search engines to easily locate me.

REQUIREMENT
Update page title from `United States Web Design System` to say `U.S. Web Design System (USWDS)` in the config.yml file. Update the description to use the current tagline.

JUSTIFICATION
Use acronym since it is the way we talk about the program
Improve organic search
Improve readability, scannability by reducing USWDS as modifier


mejiaj:
LGTM

PR #939, opened 2020-05-04
USWDS-SITE-937 Date Picker Guidance

wbyoko:
Screenshot 2020-05-04 13 31 26

Screenshot 2020-05-04 14 18 40


Ben-Judy-Flexion:
Regarding When to use the date picker component:
I suggest leaving the Scheduling paragraph as-is, it's fine. But also add something like this:

`Finding an unknown date in the past: When users need to enter a date they cannot exactly remember, and a calendar view would be helpful in identifying the correct date. Example: "I remember this event occurred on the Monday after Christmas last year... what was that date?"`

This might be an edge case, but has symmetry with the 'known date' guidance to use date input fields.

thisisdano:
@Ben-Judy-Flexion I updated your point to be a little more general: 

>  **When the day of the week is important.** When knowing the day of the week helps users choose a specific date.

PR #947, opened 2020-05-08
Banner flash test

thisisdano:
I've built out some work based on the Filament work that @maya linked in https://github.com/uswds/uswds/issues/3092#issuecomment-627595756. 

[Demo](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.app.cloud.gov/preview/uswds/uswds-site/dw-banner-flash-test/) → 

### What it is supposed to do
Essentially, what it does is build a USWDS javascript loader (`uswds-loader.js`) that's called in the `<head>` whose job it is to asynchronously load the main USWDS javascript file (like uswds.js) and provide a couple status/feature flag hooks on the `html` element:

condition | action | html class | result
--- | --- | --- | ---
start of page load | `head` element read | `html` | no change from default, banner content set to open
javascript available | loader fires. Adds loading class to `html` and starts async load of USWDS js | `html.usa-js-loading` | `usa-js-loading` hook set to hide banner content
USWDS javascript loads | loader updates html class, USWDS js initializes components | `html.usa-js-present` | CSS-based hide is removed and replaced with `hidden` attribute set by USWDS js
USWDS javascript does not load | loader times out and removes loading class | `html` | banner loses CSS-based hide and reverts to open

### Importantly: Why doesn't the demo link work?
Well mostly because I don't know JavaScript. Basically, I don't know how to put a proper path to `uswds.js.min` from `uswds-loader.js`. The way I've written it, it always looks for the file _relative to the page loaded_ which means a link that works at one level of hierarchy (`path/to/page.html`) fails at a different level (`path/to/deeper/page.html`). I assume someone who knows what they're doing could work from this to build a production-ready solution. :)

### Also...
Noting that if js doesn't load on a mobile device, the whole nav is unusable. Which is no good. Perhaps it's time to reconsider the no-js requirement? Or be _way_ more cognizant of it...


sarnikbixal (commented):
@thisisdano - I took at a look at this made sure the code was functioning well enough. I am seeing this work so I think it's in a fine state. THe uswds-loader.js is fine and needs no changes. I am not too sure about the load.js file. It has some extra properties that don't seem to get used. Is this file used elsewhere or was it taken from a template of some sort? If we aren't using those other checks we should just remove them.

js/uswds-loader.js
```diff
@@ -0,0 +1,42 @@
+var loadJS = require("./vendor/loadjs.js");
+
+// EDITABLE: path to USWDS js
+var enhancedScriptPath = "../../assets/js/vendor/uswds.min.js";
```
thisisdano, line 4: This hardcoded path is probably the source of our woes.

thisisdano:
@sarnikbixal: I don't see this working on my end. For instance, look at the preview page [here](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.app.cloud.gov/preview/uswds/uswds-site/dw-banner-flash-test/) — the loader loads but it can't find the USWDS js file and the loader (properly) reverts the page to `no-js` mode after the timeout. But it _should_ find the USWDS js. If you look at a page like [this one](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.app.cloud.gov/preview/uswds/uswds-site/dw-banner-flash-test/design-tokens/z-index/), you'll see everything working as expected, the loader loads and the USWDS js loads as well with no flash. This should be what we see on every page.

So why are some pages failing and others not? The culprit appears to be the var at line 4 in `uswds-loader.js` that I've referenced above:

```js
var enhancedScriptPath = "../../assets/js/vendor/uswds.min.js";
```

This hardcoded path only works on pages where that hardcoded path is valid. That is, pages that are a couple levels deep in the site hierarchy: The loader script is looking for the USWDS js is a location _relative_ to the current page. Our script needs to be smarter than that — it needs to be able to find the USWDS js no matter what page we're on.

So, the loader needs to be customizable so developers can point it to the proper location of their site js (like `uswds.min.js`) — _or_ it needs to be able to determine this location automatically (like always look in the same directory as `uswds-loader.js`). How can we do that?

sarnikbixal:
@thisisdano After playing around with this I realize there is no way to "find" this file dynamically without running a server process "ie during the gulp process". The simplest solution is for the user to modify that path to the uswds.min.js. The safest solution is to add a process to the gulp task at runtime to find that uswds.min.js path and set that enhancedScriptPath. I can look into that if we want to go that route.

thisisdano:
Resolved with `uswds-init` in USWDS 2.10.0

PR #978, opened 2020-06-16
Add combo box implementation guidance

thisisdano:
Added:
**Add a default value or a placeholder with data attributes.** Add a default value to the combo box by adding a `data-default-value` attribute to the `div.usa-combo-box` element using the id of the option you want to be the default value as the attribute's value. Add a placeholder to the combo box by adding a `data-placeholder` attribute to the `div.usa-combo-box` element. The placeholder value can be any string.


thisisdano:
@wbyoko Anything I'm missing here, or any changes you would make?

thisisdano:
closed in favor of https://github.com/uswds/uswds-site/pull/982

PR #987, opened 2020-06-25
USWDS: File Input Documentation

jaredcunha:
## Description

Adds documentation for file input

## Additional information

Include any of the following (as necessary): 

* Relevant research and support documents
Screen Shot 2020-06-25 at 6 50 42 PM

- [ ] Run `npm test` and make sure the tests for the files you have changed have passed.
- [ ] Title your pull request using this format: [Website] - [UI component]: Brief statement describing what this pull request solves.


_components/form-controls/10-file-input.md
```diff
@@ -0,0 +1,54 @@
...
+<div class="usa-accordion usa-accordion--bordered site-accordion-docs">
+  <button class="usa-button-unstyled usa-accordion__button"
+      aria-expanded="true" aria-controls="file-input-docs">
+    Guidance
+  </button>
+  <div id="file-input-docs" aria-hidden="false" class="usa-accordion__content site-component-usage">
+
+    <h4>When to use the file input</h4>
+    <ul class="usa-content-list">
+      <li><strong>Documents are required.</strong> You should only ask users to upload files when it is critical to service delivery.</li>
```
Ben-Judy-Flexion, line 20: You use the word "upload" here but later mention that this component doesn't upload files. Maybe use the word, attach?

_components/form-controls/10-file-input.md
```diff
@@ -0,0 +1,54 @@
...
+    </ul>
+    <h4>When to consider something else</h4>
+    <ul class="usa-content-list">
+      <li><strong>Documents are optional.</strong> Don’t ask users to provide documents if you do not require them.</li>
+      <li><strong>Upload functionality.</strong> The USWDS file input does not currently upload files.</li>
```
Ben-Judy-Flexion, line 25: This might need some clarification. I think the point we're making is that it doesn't _immediately_ upload when the user selects the file(s.) Or are we saying something different? Perhaps mention what it does do ("attaches files which will be uploaded upon form submission," or something like that.) Why would someone need upload functionality, though?

Ben-Judy-Flexion: Maybe another "when to consider something else" case is when the user is uploading very large files.

_components/form-controls/10-file-input.md
```diff
@@ -0,0 +1,54 @@
...
+    </ul>
+    <h4>Usability guidance</h4>
+    <ul class="usa-content-list">
+      <li><strong>Allow multiple file formats.</strong> Users won’t have the same software. Allow multiple types of files so user won’t need to download new software.</li>
```
Ben-Judy-Flexion, line 29: I'm not clear on this one. I get that we don't want to be overly restrictive in what file formats can be attached. I just don't understand the point about users not needing to download new software.

_components/form-controls/10-file-input.md
jaredcunha: Good point - better to be precise. I've reworded this "provide"

_components/form-controls/10-file-input.md
jaredcunha: I've changed this to say that the file input only posts files when the form is submitted. The bold text says "upload in place"

I think regarding the file size, we can say something like "asking for large files" so that we respect data limits or spotty signals.

_components/form-controls/10-file-input.md
jaredcunha: I'm wrestling with how to word this. You want to be flexible in what formats to allow. Like, don't require that files be `.docx`. Some people might use pages, which by default gives a `.pages` extension

jaredcunha:
@thisisdano This is ready for review. I have a couple of overarching questions about what to put into the guidance as well. 
- Should we mention how this component progressively enhances `<input type="file">`? 
- Should we talk about how for screen readers, the experience doesn't change?
- Worth including the note about IE and Edge?

PR #1045, opened 2020-09-10
Add identifier component guidance and add to site footer

thisisdano:
[Preview](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.app.cloud.gov/preview/uswds/uswds-site/add-identifier/components/identifier/)

This is still pending final final Spanish copy, but it's good enough to review and merge right now. We can add final copy in a subsequent PR.

### DIFF
--- a/_data/settings/components/identifier.yml
+++ b/_data/settings/components/identifier.yml
@@ -1,0 +1,27 @@
+name: Identifier
+contents:
+  - name: Background color
+    description: The background color of the identifier. Use a color of **grade 80 or higher**, darker than the section that precedes it.
+    var: $theme-identifier-background-color
+    default: '"base-darkest"'
+    type: color
+  - name: Font family
+    description: The font family of the identifier
+    var: $theme-identifier-font-family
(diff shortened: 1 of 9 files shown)


_data/settings/components/identifier.yml
```diff
@@ -0,0 +1,27 @@
...
+    type: color
+  - name: Font family
+    description: The font family of the identifier
+    var: $theme-identifier-font-family
+    default: '"ui"'
+    type: family
+  - name: Domain text color
+    description: The color of your domain text in the identifier masthead. This should be **grade 20-30** in the same family as the `$theme-identifier-background-color`.
+    var: $theme-identifier-identity-domain-color
+    default: default
```
mejiaj, line 16: Should this be `"primary-lighter"`?

thisisdano: No — we intend it to be the same family as the component background, but I see that's not how it's set up in settings. Changing... Good catch

mejiaj (approved):

PR #1060, opened 2020-09-22
Use Sass modules, simplified. [POC]

thisisdano:
**Do not merge. If we use this work, we'll set up a new feature branch.**

[Preview](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.app.cloud.gov/preview/uswds/uswds-site/dw-test-module-load/) → 

Notes:
- Uses a 2.9.0 Sass modules branch not yet merged into `uswds/use-sass-modules`.
- Ignore the `.snyk` changes (needed to make build pass)
- Ignore the `sass.js`  and `uswds-theme-typography.scss` changes (prettier)
- `*-theme-package` and `*-theme-test` files are for demonstration purposes only

Only 4 files changed that really matter:
- `uswds-components.scss` 
- `uswds-custom.scss` 
- `uswds-fonts.scss` 
- `uswds-utilities.scss` 

- - -

This PR attempts to use a hybrid approach for integrating Sass modules into the USWDS website, an approach that might be closer to what we could suggest as a first draft for devs in a similar position.

### You can use @import to use USWDS
The following entry point — which is the same as what we suggest in our [theme starter files](https://github.com/uswds/uswds/blob/develop/src/stylesheets/theme/styles.scss) — is completely valid for using USWDS with Sass modules, _and requires no other changes in settings or custom files_.

```scss
// import all settings as before
@import "uswds-theme-general";
@import "uswds-theme-typography";
@import "uswds-theme-spacing";
@import "uswds-theme-color";
@import "uswds-theme-utilities";
@import "uswds-theme-components";
// import uswds as before
@import "uswds";
// import custom-styles as before
@import "uswds-theme-custom-styles";
```

**Developers using _only_ the standard USWDS package should see no change in using the Sass modules implementation.**

### Using the benefits of Sass modules requires more work
If you want to use the module system to add namespacing or to subset the design system with packages, you need to separate the module-enhanced styles from the older import-based styles.

The following works for adding custom Sass that has access to USWDS core functions, mixins, etc — but it does not allow namespacing like `@use "uswds-core" as uswds;`:

```scss
@import "uswds-theme-general";
@import "uswds-theme-typography";
@import "uswds-theme-spacing";
@import "uswds-theme-color";
@import "uswds-theme-utilities";
@import "uswds-theme-components";
// load uswds-core package
@import "packages/uswds-core";
// load custom styles that have access to uswds-core features
@import "uswds-theme-custom-styles";
```

To do this, you need to `@forward` your custom styles files from an entry point, but you can't mix module-based rules with import-based rules. You need to create what might be called a module-based "theme package" to include from the primary entry point that serves as the module-based entry point for the subsequent work.

**1: Primary Sass entry point**
```scss
// styles.scss

@import "uswds-theme-general";
@import "uswds-theme-typography";
@import "uswds-theme-spacing";
@import "uswds-theme-color";
@import "uswds-theme-utilities";
@import "uswds-theme-components";
// load a custom styles package that is the module-based entry point for this work:
@import "uswds-theme-package-custom-styles";
```

**2: Theme package — the module-based entry point**
```scss
// uswds-theme-package-custom-styles.scss

// forward any module-based files:
@forward "uswds-theme-custom-styles";
```

**3: Stylesheet — add `@use` rules**
```scss
// uswds-theme-custom-styles.scss

// Use the uswds-core package with namespacing
@use "packages/uswds-core" as uswds;

.foo {
  color: uswds.color("red-50v");
}
```

A similar technique would work with package subsetting:

**1: Primary Sass entry point**
```scss
// styles.scss

@import "uswds-theme-general";
@import "uswds-theme-typography";
@import "uswds-theme-spacing";
@import "uswds-theme-color";
@import "uswds-theme-utilities";
@import "uswds-theme-components";
// load a custom subset package that is the module-based entry point for this work:
@import "uswds-theme-package-subset";
```

**2: Theme package — the module-based entry point**
```scss
// uswds-theme-package-subset.scss

// forward any module-based packages:
@forward "packages/usa-banner";
@forward "packages/usa-button";
@forward "packages/usa-identifier";
```

### We don't need to do too much
We could go farther with native module-based rules, like `@use`ing `uswds` with the variable overrides defined as a map, and get rid of `@import` altogether, but there's no reason to do that now as far as I can tell, and the upgrade path from non-module USWDS to module-based USWDS seems smooth and promising.


chazchumley:
@thisisdano I would love to see a demo.  Also, `@imports` will be deprecated so those would need to change to `@use` or `@forward` depending on their use.  A definite sandbox demo would allow us to talk thru a couple of scenarios.

thisisdano:
We are not using Sass modules in the near term

PR #1155, opened 2021-02-19
USWDS - Modal window guidance

jaredcunha:
## Description

Guidance for modal windows

Component PR: https://github.com/uswds/uswds/pull/3981
Issue: https://github.com/uswds/uswds/issues/3936


thisisdano:
@jaredcunha How would you trigger a modal that's not triggered by a button? For instance, a timeout modal that would be triggered by some kind of timer. Or one that's triggered by trying to close the browser window? 

Does the modal content have to be available in the DOM when the page loads, or can it be "pulled from somewhere" (technical talk) when needed. Sorry if this is obvious (or too complicated to get into here!), but I was trying to understand if all the necessary guidance is in our implementation section.

jaredcunha:
@thisisdano - The way the modals are set up, you'd have to trigger a `click()` event on a hidden button since modals require an opening element.  Here's a codepen that I written that automatically opens a modal: https://codepen.io/jaredcunha/pen/MWbZQrZ

The opener should be hidden so that it does not appear in rotor menus or receive focus. Using the `display-none` utility works in that instance. Alternatively, you can use a combination of `tabindex="-1"`, `aria-hidden="true"`, and `position: absolute; left: -9999em` is you're not using utilities. 

I think the main question here is where `focus` goes after you close the modal since it's typically transferred back to the button that opens it. I've added a test link before the hidden opener - it seems the focus is automatically transferred to the beginning of the page, which might make sense.

We had indicated this as something that required a bit more research if we had more time since there might be a better approach, but this approach seems to work given the current implementation.

jaredcunha:
Oh - BTW - I noticed that the `aria-label` on the close buttons are empty in the component code:

```html
<button class="usa-button usa-modal__close" aria-label="" data-close-modal>     
  <svg class="usa-icon" aria-hidden="true" focusable="false" role="img">
    <use xlink:href="/preview/uswds/uswds-site/accelerator-3936/modal/assets/img/sprite.svg#close"></use>
  </svg>
</button>
```

jaredcunha:
Found another issue - the `<ul>` should have `<li>` as child elements. 

```html
<ul class="usa-button-group">
            <button type="button" class="usa-button" data-close-modal>Continue without saving</button>
            <button type="button" class="usa-button usa-button--unstyled padding-105 text-center" data-close-modal>Go back</button>
</ul>
```

thisisdano:
Thanks — we'll update these asap

PR #1158, opened 2021-02-19
USWDS guidance site - add Print styles to the component section

mitchmoccia:
This PR adds Print styles to Components. The style of guidance is similar to Data visualizations and Typography since Print styles isn't an actual component. The consensus was that it should go in the Components section so it doesn't get ignored.

The name was change to Print styles and all the guidance is in sync with the guidance doc.

For the guidance site (unlike the core site), the _print.scss file was placed in the css/settings/ directory and imported with components. That file is the same as the print file pushed to the core site:
src/stylesheets/base/_print.scss

Fed link for reference:
https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.app.cloud.gov/preview/uswds/uswds-site/accelerator/3940-printer-friendly/components/print-styles/

Related PR for core site:
https://github.com/uswds/uswds/pull/4015


css/settings/_print.scss
```diff
@@ -0,0 +1,101 @@
+@media print {
```
jaredcunha, line 1: Will not be necessary since the `uswds` package will contain it?

_includes/code/components/print-styles.html
```diff
@@ -0,0 +1,101 @@
+@media print {
```
jaredcunha, line 1: Can probably remove this file too if not using the snippet in the guidance.

_data/settings/components/print-styles.yml
```diff
@@ -0,0 +1,2 @@
+name: Print styles
```
jaredcunha, line 1: If there's no data component settings, is this file needed? It shows up here: https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.app.cloud.gov/preview/uswds/uswds-site/accelerator/3940-printer-friendly/documentation/settings/#print-styles

css/uswds-components.scss
```diff
@@ -1,3 +1,4 @@
+@import 'print';
```
jaredcunha, line 1: Then if the above is deleted, this isn't required either

thisisdano:
This is going to need some more work until it's ready for primetime. Closing until we have time to pick it up again

PR #1215, opened 2021-04-22
USWDS - Box-shadow: Replace nonexistent token value

hursey013:
## Description

The [docs](https://designsystem.digital.gov/utilities/shadow/) describe the use of the utility `.shadow-0`, however using that class, or the include `@include u-shadow(0)` throws the following error:

```
Error: '0' is not a valid USWDS box-shadow token. Valid tokens: none, 1, 2, 3, 4, 5
```

This corrects the doc to use the value of `none` rather than `0`.


thisisdano (approved):

thisisdano:
Merged over unrelated CI failure

PR #1232, opened 2021-05-24
Add note about using disabled attribute

thisisdano:
We should encourage developers to use the `disabled` attribute for disabled buttons, not `usa-button--disabled`, which only gives the _appearance_ of a disabled button.


_components/button/guidance/implementation.md
```diff
@@ -1,2 +1,3 @@
 - The examples demonstrate how to use button elements. To use a button style on a link, add the `usa-button` class to your link.
-- To use a different style button on your link, add button variant classes in addition to `usa-button`.
\ No newline at end of file
+- To use a different style button on your link, add button variant classes in addition to `usa-button`.
+- Use the browser-native `disabled ` attribute for any disabled button. Don't use  `usa-button--disabled`, which is intended only for debugging and prototyping.
```
mejiaj, line 3:
```suggestion
- Use the browser-native `disabled` attribute for any disabled button. Don't use  `usa-button--disabled`, which is intended only for debugging and prototyping.
```


Currently looks like this with extra space:
image

mejiaj (approved):

PR #1259, opened 2021-08-31
This is a useful tool for USWDS Tokens

mgifford:
It displays the grids, provides the hex and has a nice search interface too.

Makes sense to me that it should be linked to from this page for others making use of the tokens.

Note, I do work for CivicActions, but didn't create this Token Color Tool.


mejiaj:
@thisisdano thoughts on adding it to [implementations page](https://designsystem.digital.gov/documentation/implementations/) instead?

mgifford:
So maybe here https://designsystem.digital.gov/design-tokens/color/theme-tokens/ ? 

I can re-roll another patch if that's where it should go. Could just be a link that describes a visual representation.

thisisdano:
OK, I think what we'll want to do here is make a `Tools` section on the color overview page, underneath `References` as another `h2` and add this link in there. The theme tokens page is probably too specific for this tool. 

We should also add both `References` and the new `Tools` sections to the sidenav (add them after line 20 in the front matter...)

mejiaj:
Added the new section and subnav elements in this new PR
https://github.com/uswds/uswds-site/pull/1276

PR #1273, opened 2021-09-23
Proposed link to CivicActions Accessibility Site

mgifford:
I think our practice area site is of value to government, particuarly with the list of open source tools/resources.

There is a range of information which is valuable for many working with digital content and it is a site that encourages interaction and engagement to more closely align with government needs.


mejiaj (approved):

thisisdano:
I think this is a valuable resource and I appreciate wanting to share it. We have to walk a fine line when we're including non-government resources on our site, and in this instance I'm going to have to say no.

PR #1435, opened 2022-02-10
Copyedit round 1 icon

mikki-stacey:
## Description
Revised for grammar, consistency, and accessibility


mikki-stacey:
@mejiaj , in the implementation file, please confirm whether uswds-gulp is correct or if we should revise to uswds-compile.

mejiaj (changes requested):
Noted reference to uswds-compile, instead of gulp.

mejiaj (approved):

thisisdano (approved):

PR #1478, opened 2022-03-16
Fix typo / inconsistency in "display-none" code example text

aduth:
## Description

Fixes a minor text inconsistency on the ["Display" utilities page](https://designsystem.digital.gov/utilities/display/#utility-display), where each `display-` utility class includes a code snippet with (a) the class name and (b) text matching that class name. However, the text for the `display-none` example is inconsistent, showing `display-inline` instead of `display-none`. This makes it marginally more difficult to identify `display-none` as an available utility, where one would need to reference the code snippet since the nature of `display-none` would make the live preview invisible.


thisisdano:
Closing in favor of https://github.com/uswds/uswds-site/pull/1630

PR #1509, opened 2022-04-05
Load components from the proper Fractal source

thisisdano:
We updated USWDS Fractal to render to the `/_site` directory instead of `/build`. This PR updates the build and component scripts to use the new directory.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -104,7 +104,6 @@ PLATFORMS
   x86_64-linux
 
 DEPENDENCIES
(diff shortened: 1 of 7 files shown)


mejiaj (changes requested):
This works if we install `uswds@latest`. We should also update the version in `_config.yml`.

https://github.com/uswds/uswds-site/blob/c9b5d98e010f2904f06ea6c296d6776cf29010de/_config.yml#L5

PR #1523, opened 2022-04-13
Use uswds@develop branch

thisisdano:
Build from develop branch to incorporate new USWDS features. Specifically, this allows us to access the explicit `@frctl/mandelbrot` dependency that's causing our circle builds to fail.


amyleadem (approved):
Looks good to me -- I get successful builds on my end. Thanks!

PR #1553, opened 2022-04-26
USWDS-site: Repair prettier command

amyleadem:
## Description
The command `npm run prettier:scss` is returning different results than its paired npx command `npx prettier --write ./css/**/*.scss`. 

Running the command via `npm` causes it to only look at directories exactly one level deep inside `/css` while running it via `npx` brings results from multiple levels. See screenshot for differences:

image

## Solution
Wrapping the path in single quotes returns intended results.
image


mejiaj (approved):
Great, this works! Thank you!!

I added some ugly CSS to `uswds-fonts.scss` and ran the previous command (nothing happened). Then I ran it with the new one and it prettified as expected.

PR #1621, opened 2022-05-29
Added online version of autoprefixer.

rahulkumarsingh73690:
Added online version of Autoprefixer. So users can instantly use it without installing it.


mejiaj:
Hey @rahulkumarsingh73690, thanks for contributing! 

While it's a valid concern, I don't think it makes sense to add this online tool in the context of the paragraph. We're talking about the `uswds-compile` tool as a method for automatically handling prefixing for you.

PR #1646, opened 2022-06-08
Update CircleCI to large resource class in config.yml

thisisdano:
CircleCi suggested I update the resource class of our containers, so I did!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,6 +103,8 @@ jobs:
       - run:
           name: Run pa11y-ci desktop
           command: npm run start-detached && npm run pa11y-ci:sitemap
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -103,6 +103,8 @@ jobs:
       - run:
           name: Run pa11y-ci desktop
           command: npm run start-detached && npm run pa11y-ci:sitemap
+    # The resource_class feature allows configuring CPU and RAM resources for each job. Different resource classes are available for different executors. https://circleci.com/docs/2.0/configuration-reference/#resourceclass
+    resource_class: large
```
thisisdano, line 107: Note that this commit was automatically generated by CircleCI. I don't care about this comment in the config, but I'm happy to delete it if it's distracting

mejiaj (approved):

PR #1673, opened 2022-06-22
Update when-to-use for unclear text

bonnieAcameron:
-Response to a report that said the "calendar picker" was confusing wording, presumably since we typically use "date picker" language
-I updated the language from "calendar picker" to "date picker"
-I hyperlinked to the "date picker" page for further clarity


mejiaj (approved):
@bonnieAcameron I've updated the link to use markdown syntax, instead of HTML. 

```diff
- <a href="https://designsystem.digital.gov/components/date-picker/">date picker</a>
+ [date picker]({{ site.baseurl }}/components/date-picker)
```

I've also updated the URL to use the site's base url. This allows us to test the link in all environments. For example, before the federalist link would point to the production site. Now it points to the component page in the same federalist preview.

Issue #1708, opened 2022-07-11
Provide more context around table


bonnieAcameron:
### Describe the bug

We received feedback in touchpoints that Phase 2 on the Getting Started documentation page has an [unclear table](https://designsystem.digital.gov/documentation/getting-started/developers/phase-two-compile/#step-4-create-path-settings-and-export-compile-functions).

User feedback: "Step 4 is woefully deficient. It does not explain where to create your path settings or why."

To fix:

- [x] add some context around path settings for this step so that users can understand where, why, and how to perform Step 4. 

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://designsystem.digital.gov/about/community/#community-conduct).
- [X] I checked the [current issues](https://github.com/uswds/uswds-site/issues) for duplicate bug reports.

bonnieAcameron:
Per meeting with Amy: 
Where is in the gulp file in step 3
Why is 

when it initializes it pulls and copies elements from noede mod package 93.0) cipies into project (that's source) and the dist is where they want it ot build. What they send up to their server. in V3, we imaigine there is an assets directory where the defualit is its ownsubdirectory that hold as all imag font cs js, but if they want to go to build or dist directory, this is where they'd do that to eliminate the uswds subdirectory

PR #1760, opened 2022-09-02
update step 4 table with context

bonnieAcameron:
Adding table context according to problem statement & notes in #1708 

Preview link: [Phase 2, Step 4 of Getting Started for Developers](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.app.cloud.gov/preview/uswds/uswds-site/bc-add-table-context/documentation/getting-started/developers/phase-two-compile/#step-4-create-path-settings-and-export-compile-functions)

### DIFF
--- a/pages/documentation/getting-started-developers/phase-two.md
+++ b/pages/documentation/getting-started-developers/phase-two.md
@@ -124,7 +124,10 @@ exports.compile = uswds.compile;
   ```
 
 ## Step 4: Create path settings and export compile functions
-Now, create your path settings using the following table:
+If you want to go to build or the dist directory and eliminate the USWDS subdirectory, create your path settings in the gulpfile. When you're initializing, the file pulls and copies elements from the node mod package of USWDS 3.0 and copies it into the project/source file while you are building in the dist file. 
+
+
+ You can build that directory using the following table:
 
 <div markdown="1" class="usa-table-container--scrollable" tabindex="0">
 


pages/documentation/getting-started-developers/phase-two.md
```diff
@@ -124,7 +124,10 @@ exports.compile = uswds.compile;
   ```
 
 ## Step 4: Create path settings and export compile functions
-Now, create your path settings using the following table:
+If you want to go to build or the dist directory and eliminate the USWDS subdirectory, create your path settings in the gulpfile. When you're initializing, the file pulls and copies elements from the node mod package of USWDS 3.0 and copies it into the project/source file while you are building in the dist file.
```
mejiaj, line 127: >If you want to go to build or the dist directory and eliminate the USWDS subdirectory

On re-read this makes sense. _As a user_ if I want to avoid including a directory in my project for USWDS SASS I would use USWDS Compile. The way you would do this is by **not** modifying `uswds.paths.src` settings.

```suggestion
If you want to use the default dist directory and eliminate the need for a project specific USWDS subdirectory, use the default `src` paths in the gulpfile. When you're initializing, the file pulls and copies elements from the node package of USWDS 3.0 and copies it into the project/source file (your defined `paths.dist` set
tings) while you are building. 
```

This will probably need more work, but I'm trying to say:

1. `paths.src` are default paths to USWDS  —  users **should not** need to modify these _unless_ they want to include USWDS dist files into their project.
2. `paths.dist` are project specific paths. Here they'll point to where they want specific assets to live. The most common use case is defining `paths.dist.theme` because this is where your SASS entry point and theme overrides will live.

Maybe we should remove the `paths.src` settings from the table and only show `dist` settings (project settings) with a link to the full table?

mejiaj (changes requested):
I like that we're looking at improving this section!

If we can clarify further the difference between src and dist path settings and what it means to users  —  that would be a nice improvement.  Added some notes based on what _my_ interpretation was from a user standpoint.

amyleadem:
Closing in favor of https://github.com/uswds/uswds-site/pull/1766

PR #1786, opened 2022-09-27
[Test] Touchpoints

amyleadem:
Preview link: https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-touchpoints-refer/how-to-use-uswds/

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -11,7 +11,7 @@ meta:
   og:image: /img/uswds-logo/lg-black.png
 touchpoints:
   active: true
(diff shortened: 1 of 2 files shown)


amyleadem:
Referrer works via modal submission:
image

Need to check if we still encounter the errors detailed in Issue https://github.com/uswds/uswds-site/pull/1541 that caused us to switch delivery methods in the first place.

Issue #1754, opened 2022-08-29
USWDS-Site: Investigate solutions for Touchpoints not showing referrer


bonnieAcameron:
I've found [this](https://developer.chrome.com/blog/referrer-policy-new-chrome-default) so far, which points to CSP policies.

![uswds site](https://user-images.githubusercontent.com/96838068/187274393-9f938c65-2264-47f3-8be3-428dc5fbeb9c.png)

And the site shows:

![referrer](https://user-images.githubusercontent.com/96838068/187274420-951b98a3-05c0-4858-b79d-d3d579e03dd1.png)

PR #1788, opened 2022-09-27
USWDS-Site: Turn on Touchpoints modal

amyleadem:
## Summary
Updated the Touchpoints markup to use the "Custom button & modal" delivery method in our production environment and the "Hosted only on the Touchpoints site" delivery method in all other environments. 

The modal delivery method enables Touchpoints to receive accurate data in the `Page submitted from` field. However, the related modal script caused an increase in false positives and page crashes in pa11y/circleci. For this reason, we decided to implement the modal only in production, where it would not affect our pa11y tests. 

## Related issue
Closes https://github.com/uswds/uswds-site/issues/1754

## Preview link
Preview link: [Sample page](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-touchpoints-modal/how-to-use-uswds/)

## Problem statement
The "Was this page helpful?" survey results from Touchpoints currently do not return helpful content in the "Page submitted from" metadata field. Instead, it simply reports the location of the survey page for every entry: 
```
Page submitted from
/touchpoints/344131bb/submit 
```

## Solution
During testing, it was revealed that using the modal delivery method resulted in useful location data output. PR https://github.com/uswds/uswds-site/pull/1786 revealed that "Page submitted from" data is accurate when submitted via modal:

image
([Example survey result report](https://touchpoints.app.cloud.gov/admin/forms/344131bb/submissions/1518195))

However, using the modal introduces an increase in false positives and page crashes in pa11y, as outlined in https://github.com/uswds/uswds-site/pull/1541. 

To resolve this, we limited the use of the modal to the production environment only. This will allow us to gather the results we need from the live site while also preventing additional pa11y/circleci errors. 

---

Before opening this PR, make sure you’ve done whichever of these applies to you:

- [x] Run `git pull origin [base branch]` to pull in the most recent updates from your base and check for merge conflicts. (Often, the base branch is `develop`).
- [x] Run `npm test` and confirm that all tests pass.


amyleadem:
## Failure notes
[7878845](https://github.com/uswds/uswds-site/pull/1788/commits/7878845d99ab0cdb6a263628ed906211b3e13395) - accessibility failures

_config.yml
```diff
@@ -9,7 +9,6 @@ federalist_component_preview: "iframe.html?id="
 federalist_release_prefix: "release-"
 touchpoints:
   active: true
-  hosted: true
 # this is for pages that don't have a permalink (primarily posts)
```
mejiaj, line 12: Does this remove the option of toggling the hosted version in different environments?

mejiaj (commented):
Looks good, just have a question on the hosted option. Whether it will affect our ability to toggle hosted on production (for testing or any other case).

_config.yml
amyleadem: @mejiaj It does. The `hosted` option does not provide the necessary `source` data, so it is now hard-cored that the production environments have the equivalent of what was `hosted: false` and the non-production environments have the equivalent of `hosted: true`. I felt it could introduce error and confusion to retain this as a configurable option, so I removed it. 

Let me know if you think it would be helpful to keep this config.

mejiaj (approved):

PR #1856, opened 2022-10-25
[IDP] Input mask component guidance

mitchmoccia:
This contains the latest guidance for the input mask component.


thisisdano:
Merging this over the Circle fails:
- `test_build` will fail because of links to patterns that aren't yet in the branch
- `test_a11y_mobile` failed because of a puppeteer crash, but the in-page nav a11y tests passed

PR #1896, opened 2022-11-10
Idp staging in-page nav fix which removes in-page nav from landing pages

mitchmoccia:
I created a new layout called in-page-nav that has the code for in-page nav and REMOVED the in-page nav code from the styleguide layout. Then I changed component.html to call `layout: in-page-nav`.

The result is landing pages don't have in-page nav. 

Why? Team consensus was that the landing pages don't need it. There were multiple issues, including three card layouts with h3 tags on the same row, and three links in the nav which would all point to the same vertical scroll-to position.

I went through the patterns pages too, and basically **any page that has a card layout (e.g. n number of rows with three cards per row) I used the standard styleguide layout.

### DIFF
--- a/_components/button-group/button-group.md
+++ b/_components/button-group/button-group.md
@@ -4,7 +4,7 @@ component:
   name: button-group
   status: ready
   package: usa-button-group
-layout: styleguide
+layout: in-page-nav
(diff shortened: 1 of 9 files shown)


thisisdano:
We're going to do this differently

PR #1900, opened 2022-11-11
IDP component alert and combobox content updates

jonathanbobel:
# Component Alerts and Combobox content updates

## Description

Added the capability to conditionally add an alert box to the top of a component page. To use, set alert:true and add the following to the component's front matter:
- alert-class
- alert-header
- alert-content

Additionally, updated the guidance to include accessibility concerns and a link to the github issue detailing them


thisisdano (approved):
There were some weird merge conflicts and merge artifacts in this PR, but I think I've addressed them all. I updated the alert language a bit to be clear about when we found this issue and our next steps

thisisdano:
The final PR is good to go and is properly scoped to only changes associated with the combo box alert

PR #1907, opened 2022-11-17
Add changelog entry for Memorable Date updates

thisisdano:
This adds a changelog entry to Memorable Date. We updated the month entry to a `select` from a numeric `input` because our pattern testing showed that a `select` was clearer and reduced confusion. 

Note that this is not a breaking change even though it affects styles and markup. We rewrote the styles to be backward compatible with the old markup


jonadecker (approved):
💯

PR #1949, opened 2022-12-16
USWDS-Site - Spacing token - Repair converted tokens

### LINKED ISSUE #1934: USWDS-Site - Spacing tokens: Incorrect value for token `0.5`
### Describe the bug

On the spacing tokens page, the token `0.5` appears to be falsely linked to token `5`. `units(0.5)` results in a value of `.25rem`, which is the equivalent of token `'05'`.

https://designsystem.digital.gov/design-tokens/spacing-units/

image

### Steps to reproduce the bug

_No response_

### Expected Behavior

_No response_

### Screenshots

_No response_

### System setup

_No response_

### Additional context

_No response_

### Code of Conduct

- [X] I checked the [current issues](https://github.com/uswds/uswds-site/issues) for duplicate bug reports.

amyleadem:
## Related issue
Closes #1934

## Preview links
- [Tokens - Spacing units](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-spacing-token-05/design-tokens/spacing-units/)
- [Tokens - Flex](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-spacing-token-05/design-tokens/flex/)
- [Tokens - Order](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-spacing-token-05/design-tokens/order/)
- [Tokens - Shadow](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-spacing-token-05/design-tokens/shadow/)

## Problem statement
On the spacing tokens page, the token `5` is incorrectly linked to the `0.5` number equivalent. `0.5` should instead be linked to token `'05'`.  This appears to happen because of confusion with the leading zero and quotes in `'05'` &ndash; something is getting lost in translation between the files that prevents `'05'` from receiving any `converted` value. `5` receives the converted value instead. 

image

On the spacing unit page, we need to both:
1. Prevent the `5` token from receiving the `0.5` number equivalent AND 
1. Associate the `'05'` token with the `0.5` number equivalent

Additionally, the  following pages also need to break the association between the `5` token and the `0.5` conversion:
- [Order tokens](https://designsystem.digital.gov/design-tokens/order/)
- [Flex tokens](https://designsystem.digital.gov/design-tokens/flex/)
- [Shadow tokens](https://designsystem.digital.gov/design-tokens/shadow/)

## Solution
The `converted` assignment in `is_number.html` consistently ignored the leading zero in the `'05'` token and would assign the `converted` value to `5` instead. Attempts to explicitly declare `'05'` as a string to force the leading zero were not successful.

As a workaround, this PR creates an explicit `conversion` value for `'05'`  in `spacing.yml`. Declaring an explicit value here removes the confusion that is currently happening between files. 

The spacing unit page appears to be the only instance of `'05'` token that is problematic.

## Screenshots
Final result:

image

## Testing and review
To review:
- Confirm that the `5` token does not have the `0.5` number equivalent on any of the preview pages.
- Confirm that the `'05'` has the `0.5` equivalent on the spacing tokens page.
- Confirm that all pages that use `{{ converted }}` display correct values.


_data/tokens/conversion.yml
```diff
@@ -32,8 +32,6 @@
   number: -2px
 - token: neg-1px
   number: -1px
-- token: '05'
-  number: 0.5
```
amyleadem, line 36: Removing this because it causes `5` to be linked to `0.5`

amyleadem:
@mejiaj This PR creates a workaround for the confusion that happens with the leading zero and quotes between files. I tried to prevent the confusion with different methods of making `'05'` an explicit string in yaml, but I can't seem to make it all work. If you see alternate methods, please let me know!

mejiaj (commented):
I can see the benefit of having the ability to override a conversion. We could also approach this from another angle.

The issue seems to be `is_number.html` and how it treats the `05` value:
https://github.com/uswds/uswds-site/blob/aa4c540b646af56c01250e1dc5f3ef8fb2191446/_includes/tokens/is_number.html#L22-L30

We could add an additional check that would give us the number or the converted value, instead of both (because we're already checking in the template):

```liquid
{% if is_number %}
  {% assign converted = false %}
{% else %}
  {% assign conversion =
    site.data.tokens.conversion
    | where: 'token', item.token %}

  {% if conversion[0].number %}
    {% assign converted = conversion[0].number %}
  {% endif %}
{% endif %}
```

This resulted in a table that no longer has `0.5` as a value of `5`:
image

What do you think @amyleadem @thisisdano?

amyleadem:
@mejiaj I attempted to implement your suggested solution. It fixed the false association with the `5` token, but I still had trouble getting the system to recognize that  `0.5` should be paired with `05`. Any tips? 

image

mejiaj:
> @mejiaj I attempted to implement your suggested solution. It fixed the false association with the `5` token, but I still had trouble getting the system to recognize that `0.5` should be paired with `05`. Any tips?
> 
> image

I _think_ that's what `_data/tokens/conversion.yml` was intended for. That might be why `'05'` was included originally. There's definitely something strange going on with the logic in `is_number.html`.

It looks like `'05'` is not converted and not a number, based on the logic in template:

```
// spacing-units.md
{% if converted %}
  <code>{{ converted }}</code>
{% endif %}
{% if is_number %}
  <code>{{ item.token }}</code>
{% else %}
  // '05' returns false for both above and just prints out as token
  <code>'{{ item.token }}'</code> 
{% endif %}
```

_data/tokens/spacing.yml
```diff
@@ -7,6 +7,7 @@ positive:
   small:
     - token: '05'
       value: 4px
+      conversion: 0.5
```
amyleadem, line 10: Created an explicit equivalent number value here because it could not recognize the `'05'` token when assigning `converted` in `is_number.html`

_data/changelogs/tokens-spacing.yml
```diff
@@ -2,6 +2,19 @@ title: Spacing tokens
 type: token
 changelogURL:
 items:
+  - date: 2022-12-19
```
amyleadem, line 5: Need to update this date prior to merge

amyleadem:
@mejiaj 
I incorporated your logic to check for `is_number` before assigning `converted`. Thanks for that suggestion.

I was not able to get `converted` to recognize the leading zero in `'05'`. For this reason, I believe the best path forward is to set an explicit conversion value, which is found in this PR. Let me know if you have any questions or concerns.

Also - the checks are currently failing because snyk ignore needs to be updated. The forthcoming snyk fix is in this PR: https://github.com/uswds/uswds-site/pull/2013

pages/design-tokens/spacing-units.md
```diff
@@ -56,6 +56,8 @@ Any spacing value in utilities or component CSS should use the following spacing
             <span class="text-no-wrap padding-right-4">
               {% if converted %}
                 <code>{{ converted }}</code>,
+              {% elsif item.conversion %}
```
mejiaj, line 59: I would just suggest we add a note for what this means. That way we can understand the logic flow in the future.

mejiaj (commented):
Nice work fixing this issue! The output table on the spacing units page looks good.

Just a minor question on the PR date in changelog. Otherwise LGTM!

_data/changelogs/tokens-spacing.yml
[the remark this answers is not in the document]
amyleadem: Updated in [ed3e647](https://github.com/uswds/uswds-site/pull/1949/commits/ed3e64744df83e3bcf759aa9ae7225d34b49db6c)

pages/design-tokens/spacing-units.md
amyleadem: Added comments in [b4b7ad7](https://github.com/uswds/uswds-site/pull/1949/commits/b4b7ad7227d2f89167ba90e951cb5dcec3990b00)

mejiaj (approved):

PR #1942, opened 2022-12-08
Deleted the what's on this page navigation in the body of the page

bonnieAcameron:
## Description

This issue affects the Contribute page as described in #1935 . I've removed the section "What's on this page" and all of the anchor links beneath it. 

## Additional information

This is a result of a recent change to uswds-site that added the side-nav component to our documentation. The end result for the Contribute page was that there were 3 navigations on a single document. This change will lighten the visual load and make the remaining page navigation more meaningful. 

B35AFC49-F613-4A94-A79C-CE03721EC372

Before you hit Submit, make sure you’ve done whichever of these applies to you:

- [ ] Follow the [18F Front End Coding Style Guide](https://pages.18f.gov/frontend/) and [Accessibility Guide](https://pages.18f.gov/accessibility/checklist/).
- [ ] Run `npm test` and make sure the tests for the files you have changed have passed.
- [ ] Run your code through [HTML_CodeSniffer](http://squizlabs.github.io/HTML_CodeSniffer/) and make sure it’s error free.
- [ ] Title your pull request using this format: [Website] - [UI component]: Brief statement describing what this pull request solves.

PR #2000, opened 2023-02-09
Remove duplicative page navigation

bonnieAcameron:
## Description

This issue removes duplicative in-page navigation that repeats what is automatically pulled in to the right-side page navigation. There may be others on the site, but these were found in #1942 . 

## Additional information

Pages updated: 
- [Key benefits](https://designsystem.digital.gov/about/key-benefits/)
- [Website policies and notices](https://designsystem.digital.gov/about/website-policies-notices/)


amyleadem (approved):
Looks good to me! Thanks for fixing this.

PR #2046, opened 2023-04-03
USWDS-Site - Touchpoints Button: Fix unclickable links on widescreen

sknep:
# Summary

Hey yall, i noticed that the sticky touchpoints container on widescreen actually covers the content to the left and makes it unclickable. This should make the button still work and not cover the content with an invisible container that blocks click interactions.

Screen Shot 2023-04-03 at 1 06 12 PM

### DIFF
--- a/css/_uswds-theme-custom-styles.scss
+++ b/css/_uswds-theme-custom-styles.scss
@@ -420,6 +420,7 @@ img {
     bottom: 0;
     position: sticky;
     z-index: z-index(500);
+    pointer-events: none;
   }
 
   .usa-button {
@@ -427,6 +428,9 @@ img {
     @include at-media("desktop") {
       @include u-bg($site-background-color);
       margin: units(2) units(2) units(4);
+    }
+    @include at-media("widescreen") {
+      pointer-events: all;
     }
   }
 }


css/_uswds-theme-custom-styles.scss
```diff
@@ -428,6 +429,9 @@ img {
       @include u-bg($site-background-color);
       margin: units(2) units(2) units(4);
     }
+    @include at-media("widescreen") {
+      pointer-events: all;
+    }
```
mahoneycm, line 434: MDN Docs labels `pointer-events: all` as an experimental, SVG-only value. While it _does_ seem to work here. I wonder if we should change it to `pointer-events: auto` as a more consistent option to if we never set it's parent element to `pointer-events: none`

> **Note**
> `pointer-events: auto`  The element behaves as it would if the pointer-events property were not specified. [^1]

[^1]: https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events

css/_uswds-theme-custom-styles.scss
```diff
@@ -420,6 +420,7 @@ img {
     bottom: 0;
     position: sticky;
     z-index: z-index(500);
+    pointer-events: none;
```
mahoneycm, line 423: Is there any value in moving these `pointer-event` rules out of the media queries and into the base style definitions? Would we ever want them to _not_ be present?

mahoneycm (commented):
This seems like an effective and simple solution to the touchpoints container blocking click events! 

An alternative solution would be to change the way we place the button, which seems like a greater lift without any additional benefit than this PR provides.

Just two considerations on the best way to implement the pointer-events rules

css/_uswds-theme-custom-styles.scss
sknep: I picked the media query because in general, I think it's unwise to change the native click behaviors anywhere you don't _have_ to, and this limited the impact to just the screen width range where the sticky positioning causes the click-blocking overlap, and it doesn't affect native behaviors with normal positioning at all the other screen widths.

mahoneycm: Thanks for clearing that up! This is a good point and I'm happy to leave it where it is

css/_uswds-theme-custom-styles.scss
mahoneycm: Resolved in [a04c3eb](https://github.com/uswds/uswds-site/pull/2046/commits/a04c3ebad6a91408925ae830264cc72a3f66184d)

mahoneycm (approved):
Updated `pointer-events` from all to auto and merged in `main`.

Ready for your approval @amyleadem @mejiaj 👍

amyleadem (approved):
Simple and effective solution! Thank you @sknep for this submission.

mahoneycm:
@thisisdano Ready for final review

thisisdano (approved):
Thank you!

PR #2053, opened 2023-04-06
USWDS-Site - Header: Update changelog for #5008 with note about possible break

### LINKED ISSUE #2051: USWDS-Site: Add cautionary note to header changelog #5008
Add cautionary note to the header changelog for [PR #5008 ](https://github.com/uswds/uswds/pull/5008)(and maybe the release notes) so that it lets users know that if they use the extended header, the [hard-coded width of 50%](https://github.com/uswds/uswds/pull/5008/files#diff-3e633cbc98f96de92db8b6117d94e95684aa90ac9494ad898d78eff505edc63dL234) has shifted to a [default setting value of 33%](https://github.com/uswds/uswds/blob/develop/packages/uswds-core/src/styles/settings/_settings-components.scss#L94). They might see a break because of this.

amyleadem:
# Summary

Updated the changelog entry for [ USWDS PR #5008](https://github.com/uswds/uswds/pull/5008) so that it warns users of a  potential breaking change.

The break was originally flagged in Slack: https://gsa-tts.slack.com/archives/C3F14AHSQ/p1680208085858799

I have updated the PR description with the same information. We should also consider updating the 3.4.1 release notes. 

## Related issue

Closes #2051 

## Preview link

Preview link: [Header changelog](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-header-changelog-5008/components/header/#changelog)

## Testing and review
- Confirm that information accurately reflects and resolves the potential break
- Confirm there are no issues with grammar or punctuation

### DIFF
--- a/_data/changelogs/component-header.yml
+++ b/_data/changelogs/component-header.yml
@@ -3,9 +3,11 @@ type: component
 changelogURL:
 items:
   - date: 2023-03-13
-    summary: Logo text width setting now works as expected.
-    summaryAdditional: Now the design system respects the value passed to `$theme-header-logo-text-width`.
-    isBreaking: false
+    summary: Extended headers now respect the value passed to `$theme-header-logo-text-width`.
+    summaryAdditional: The default setting value might differ from the previous value.
+      This can cause unexpected line breaks.
+      When updating, confirm that there are no visual regressions.
+    isBreaking: trues
     affectsStyles: true
     githubPr: 5008
     githubRepo: uswds
--- a/_data/changelogs/component-header.yml
+++ b/_data/changelogs/component-header.yml
@@ -4,9 +4,8 @@ changelogURL:
 items:
   - date: 2023-03-13
     summary: Extended headers now respect the value passed to `$theme-header-logo-text-width`.
-    summaryAdditional: The default setting value might differ from the previous value.
-      This can cause unexpected line breaks.
-      When updating, confirm that there are no visual regressions.
+    summaryAdditional: For extended headers, the default setting value is narrower than
+      the previous hard-coded value, which can cause unexpected line breaks.
     isBreaking: true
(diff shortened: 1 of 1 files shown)


_data/changelogs/component-header.yml
```diff
@@ -3,9 +3,10 @@ type: component
 changelogURL:
 items:
   - date: 2023-03-13
-    summary: Logo text width setting now works as expected.
-    summaryAdditional: Now the design system respects the value passed to `$theme-header-logo-text-width`.
-    isBreaking: false
+    summary: Extended headers now respect the value passed to `$theme-header-logo-text-width`.
+    summaryAdditional: For extended headers, the default setting value is narrower than
+      the previous hard-coded value, which can cause unexpected line breaks.
```
mahoneycm, line 8: Would it be a good idea to specify the linebreak occurs within `usa-navbar` or on logo text? or do you think it'd be redundant with the setting name? Just considering where people might look to resolve this with CSS

```suggestion
    summaryAdditional: For extended headers, the default setting value is narrower than
      the previous hard-coded value, which can cause unexpected line breaks for `usa-logo__text`.
```

mahoneycm (approved):
Fix works as expected and I think the updated changelog listing looks good! 👍 

One small comment about clarifying where line breaks would occur but it may be unnecessary given the setting name.

_data/changelogs/component-header.yml
amyleadem: Good suggestion! Added in [e1d5aad](https://github.com/uswds/uswds-site/pull/2053/commits/e1d5aadb77ce280eb683ae9ad4b22cdaba18adc7)

mejiaj (approved):

Issue #2055, opened 2023-04-07
USWDS-Site - Feature: Create "copy code" button for Component Code examples


mahoneycm:
### Is your feature request related to a problem? Please describe.

When copying component code from our website, users must click and drag or ctrl + click. This leaves much to be desired from a user experience perspective!

### Describe the solution you'd like

Copy Code buttons are pretty common on code examples throughout the web. Our site would benefit from utilizing this feature, especially on long code blocks

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://designsystem.digital.gov/about/community/#community-conduct).
- [X] I checked the [current issues](https://github.com/uswds/uswds-site/issues) for duplicate feature requests.

PR #2078, opened 2023-05-01
Copy component refactor

mejiaj:
# Summary

Reorganized copy code component JS and added code comments.

## Related issue

Related to #2055.
 

## Preview link

Preview link: [Button](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/jm-copy-component-btn/components/button/)

The previous JS is good, but we can improve further by adding comments. We can also increase readability by separating the code out more.

### Previous

```bash
. copy-code
├── buildHTML
└── btnFunction
```

### Proposed

```bash
. copy-code
├── createWrapper
├── createCopyButton
├── copyOnClick
└── init
```

## Solution

This PR does the following:
1. Creates additional top-level `const` for further code reuse
2. Uses BEM naming convention _ex: `usa-copy-code__btn`_ (previously `usa-copy-code--btn`).
3. Separates button and wrapper creation
4. Adds JSDoc comments
5. Adds missing attribute to button
6. Adds an `init()` that creates DOM, appends elements, and adds behavior.

## Major changes

Major changes were done in JS file. Minor changes to styling to update classes.

## Testing and review

_Share recommended methods for reviewing this change._

### DIFF
--- a/css/custom-styles/_copy-code-btn.scss
+++ b/css/custom-styles/_copy-code-btn.scss
@@ -11,8 +11,9 @@
     display: none;
   }
 
-  &--button {
+  &__button {
     @include u-padding-left(2);
+    min-width: units(15);
     position: sticky;
     top: units(1);
     z-index: 100;
@@ -30,13 +31,12 @@
     }
   }
 
-  &--wrapper {
+  &__wrapper {
     display: none;
 
     @include at-media("tablet") {
       display: block;
       margin-left: units(2);
-      width: units(10);
     }
   }
 }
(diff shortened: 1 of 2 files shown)


mahoneycm (approved):
Thanks for taking this on! 

- [x] Button looks and acts as expected
- [x] The button `min-width` Is a better solution to keeping consistent button sizing between states 
- [x] Not present on mobile like before
- [x] Screen reader reads the button and success states correctly
- [x] Code is more in line with other USWDS JS files

I took some notes on the differences to better understand how to format our code moving forward! 👍 

Thanks James!

PR #2213, opened 2023-07-25
USWDS-site - [File input] - Updating page to make clear what is a variant of component vs. part of component

sarah-sch:
- Please eliminate text from USWDS repo (component library) duplicating new h3s  
- Please confirm that bold "Input has an error" text is not part of component before deleting

# Summary

- Added H3 headings to make clear which text is part of file input component and which text is a label. 
- We still need the dev team's help to eliminate duplicate text from the page (the component library) and ensure that "input has an error" text is ok to use as H3 (i.e., should it be part of the component?)

## Related issue

Closes [#5211](https://github.com/orgs/uswds/projects/8/views/17?pane=issue&itemId=24415736)

## Preview link

Preview link:
NA

## Problem statement

Body text being used to label file input component may confuse users because it seems like it might be part of the component. To fix this problem, avoid future confusion, and introduce more consistency among the component pages, we are changing that text to an h3 format. As stated above, we have added the appropriate h3s, but we need the dev team to eliminate the body text duplicating the new h3s from the component li brary.

## Solution

Using h3s instead of body text to convey name of component variations is the simplest way to identify component variation names as separate from the component itself while conveying key information to users. **NOTE** that we need the dev team to confirm that the bold "Input has an error" text on the last component example is not actually supposed to be part of the component; if it is, we will need to come up with some new text to use below the h3 so that the full component is still illustrated.

## Testing and review

After you have finished the changes above, please show content team for approval.


_includes/code/components/file-input.html
mejiaj, line 1: I've added the `site-preview-heading` class to make it consistent to other preview headers.

image


Example: [Checkbox](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/ss-update-file-input-headings/components/checkbox/)

mejiaj (approved):

mejiaj (approved):
Thanks for creating this PR. Zero a11y issues on [File input preview →](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/ss-update-file-input-headings/components/file-input/)

PR #2351, opened 2023-11-13
USWDS-Site - Snyk: Add postcss override to package.json

amyleadem:
# Summary

Added an override for postcss in the `package.json` file to prevent a snyk error.

> **Note**
> This override should be removed once the vulnerability can be resolved in @uswds/compile v1.1.0

## Problem statement
Running `npx snyk test` results in the following errror: 
```
Issues with no direct upgrade or patch:
  ✗ Improper Input Validation [Medium Severity][https://security.snyk.io/vuln/SNYK-JS-POSTCSS-5926692] in postcss@8.4.19
    introduced by @uswds/compile@1.0.0 > gulp-sourcemaps@3.0.0 > @gulp-sourcemaps/identity-map@2.0.1 > postcss@7.0.39 and 1 other path(s)
  This issue was fixed in versions: 8.4.31
```

This is a vulnerability related to a subdependency of [gulp-sourcemaps](https://www.npmjs.com/package/gulp-sourcemaps), which has has not been updated for 3 years. Because of the  dependency's slow update cycle, it is unlikely that this vulnerability will be resolved by `gulp-sourcemaps`.

## Solution
Creating a temporary override for this sub-dependency will resolve the snyk error while we wait for an update to uswds/compile. 

Alternatively, we could update the snyk ignore, but that would require periodic updates every 30 days and would not resolve the vulnerability. 

## Testing and review
1. Confirm that there are no snyk errors related to postcss
2. Confirm that creating an override is the best method for resolving this error


package.json
```diff
@@ -96,7 +96,8 @@
     "vinyl-source-stream": "^2.0.0"
   },
   "overrides": {
-    "glob-parent": "6.0.2"
+    "glob-parent": "6.0.2",
+    "postcss": "^8.4.31"
   },
```
amyleadem, line 101: There is not a standard method for adding comments to a `package.json` file, but some users have created a convention like this to add comments. Is this anything we are interested in exploring? 
```suggestion
  },
  "overridesComments": {
    "postcss": "This override should be removed once the related snyk error is resolved in @uswds/compile v1.1.0"
  },
```

package.json
```diff
@@ -96,7 +96,8 @@
     "vinyl-source-stream": "^2.0.0"
   },
   "overrides": {
-    "glob-parent": "6.0.2"
+    "glob-parent": "6.0.2",
+    "postcss": "^8.4.31"
```
amyleadem, line 100: > **Note**
> I tried to reference a value of `$postcss` as described in the [npmjs overrides docs](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides), but could not get it to recognize the reference. Not sure what I am doing wrong. It should be ok to move forward without the reference since we are viewing this override as a temporary fix and we hopefully won't have to update the postcss devDependency version before that, but might be a good thing to figure out for future overrides like this. 

```suggestion
    "postcss": "$postcss"
```

package.json
mejiaj: I like the idea of adding comments close to the code so we don't forget.  A new data field might not be necessary if we reference github issues in these types of commits. 

**My proposal**
1. Create issue for updating and removing override. 
2. Schedule issue two sprints out.
3. Add override and reference new issue in git commit message. Provides a nice reference in git blame.

**Git commit message**
```diff
- Add postcss override
+ Temporary postcss override. Remove with #ISSUE_NO
```

**Mockup**

You can see in the screenshot below. There's a line via git blame (light gray) and on hover we'll get a link to the issue.

image

mejiaj (approved):
Nice fix. No issues found on install.

thisisdano (approved):

mahoneycm (approved):
Installs correctly and contains `0` vulnerabilities 👍

PR #2356, opened 2023-11-15
USWDS-Site - Changelog:  Header compilation error [#5624]

mahoneycm:
# Summary

Adds changelog for uswds/uswds#5616

## Related PR

uswds/uswds#5616

## Preview link

- [Header changelog](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/cm-changelog-5617/components/header/#changelog)
- [Settings changelog](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/cm-changelog-5617/documentation/settings/#changelog)


_data/changelogs/component-header.yml
```diff
@@ -3,7 +3,7 @@ type: component
 changelogURL:
 items:
   - date: NNNN-NN-NN
-    summary: Resolved a compilation error when `$theme-max-header-width` is set to non-breakpoint values.
+    summary: Fixed a bug that prevented `$theme-max-header-width` from accepting a value of "none".
```
mahoneycm, line 6: Now that the test has expanded to check for non-breakpoint values, should we include that in the changelog?

amyleadem: Really, this PR doesn't touch `$theme-header-MAX-width` other than removing it from the Safari fix entirely. The root issue is that the calculation in the Safari bug fix caused the a value of "none" on the setting o throw an error. So by not using the max-width setting in the Safari fix, that specific bug is removed. I think that should be the focus of the changelog. 

Regarding `$theme-header-MIN-width`, I don't think there is any need to get into the details. Nothing is really different other than the Safari bug won't be applied to values that aren't acceptable to the setting anyway. That piece feels more like a behind-the-scenes thing.

Curious to hear what you think!

mahoneycm: @amyleadem Oh I think that makes sense. The focus on the existing bug is a good move! 👍

Issue #2042, opened 2023-03-29
USWDS-Site - Code audit: Fix inaccuracies in tokens pages


amyleadem:
## Tasks
Fix the issues on the tokens pages found during the code audit. 

## Audit findings
[Code documentation audit](https://docs.google.com/spreadsheets/d/1YNid-FkaYHKJA7ezsJpCFhycACCpMPKmiNbo2QIf7lM/edit#gid=421652741) (Google sheets) :lock:

PR #2386, opened 2023-12-04
USWDS-Site: Fix token pages from audit (Part 4)

amyleadem:
# Summary

Updated component pages according to the findings reported in the [code/developer content audit](https://docs.google.com/spreadsheets/d/1YNid-FkaYHKJA7ezsJpCFhycACCpMPKmiNbo2QIf7lM/edit#gid=421652741) (Google Docs :lock:)

> [!NOTE]
> To keep PR reviews manageable, this PR updates only some of the tokens pages. All other token page updates will be addressed in other PRs. 

> [!WARNING]
> We will likely need to update the changelog dates before merge. 

## Related issue

Related to #2042

## Preview link

Order

## Solution
Update code according to the findings in  [code/developer content audit](https://docs.google.com/spreadsheets/d/1YNid-FkaYHKJA7ezsJpCFhycACCpMPKmiNbo2QIf7lM/edit#gid=987418701) (Google Docs :lock:). 

Please note that all items highlighted yellow in the spreadsheet will be investigated/discussed in the last phase of the component page fixes. 

## Testing and review

- Confirm that all updates are accurate and make sense 
- Confirm that all _meaningful_ updates have a related changelog
    - "Meaningful" updates here mostly applied to adding or removing documentation about variants, dependencies, etc. Copy tweaks to add clarity were not considered meaningful, but a case could be made for them to be included.
- Confirm grammar and spelling for all updates

### DIFF
--- a/pages/design-tokens/flex.md
+++ b/pages/design-tokens/flex.md
@@ -126,7 +126,7 @@ Your context and coding style determine how you access USWDS flex tokens in code
         </td>
         <td data-title="Example">
           <span>
-            $theme-nav-column-flex: <code>4</code>
+            $nav-column-flex: <code>4</code>
           </span>
         </td>
       </tr>
--- a/_data/tokens/flex.yml
+++ b/_data/tokens/flex.yml
@@ -1,28 +1,28 @@
 - token: 1
-  value: '1 0 0'
+  value: '1 1 0%'
 - token: 2
-  value: '2 0 0'
+  value: '2 1 0%'
 - token: 3
-  value: '3 0 0'
+  value: '3 1 0%'
 - token: 4
-  value: '4 0 0'
+  value: '4 1 0%'
 - token: 5
-  value: '5 0 0'
+  value: '5 1 0%'
 - token: 6
-  value: '6 0 0'
+  value: '6 1 0%'
 - token: 7
-  value: '7 0 0'
+  value: '7 1 0%'
 - token: 8
-  value: '8 0 0'
+  value: '8 1 0%'
 - token: 9
-  value: '9 0 0'
+  value: '9 1 0%'
 - token: 10
-  value: '10 0 0'
+  value: '10 1 0%'
 - token: 11
-  value: '11 0 0'
+  value: '11 1 0%'
 - token: 12
-  value: '12 0 0'
+  value: '12 1 0%'
 - token: auto
-  value: auto
+  value: '0 1 auto'
 - token: fill
-  value: '1 0 0'
+  value: '1 1 0%'
--- a/_data/tokens/order.yml
+++ b/_data/tokens/order.yml
@@ -1,7 +1,7 @@
 - token: initial
   value: initial
 - token: first
-  value: first
+  value: -1
 - token: 1
   value: 1
 - token: 2
@@ -25,4 +25,4 @@
 - token: 11
   value: 11
 - token: last
-  value: last
+  value: 999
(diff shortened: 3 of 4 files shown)


_data/changelogs/tokens-order.yml
```diff
@@ -0,0 +1,9 @@
+title: Order tokens
+type: token
+changelogURL:
+items:
+  - date: 2023-12-05
```
amyleadem, line 5: > [!important]
> We need to update the changelog dates before merge

pages/design-tokens/opacity.md
```diff
@@ -105,21 +105,6 @@ Your context and coding style determine how you access USWDS opacity tokens in c
...
-          <span>
-            <a href="{{ site.baseurl }}/design-tokens/opacity/" class="token">opacity</a>
-          </span>
-        </td>
-        <td data-title="Example">
-          <span>
-            $theme-hover-opacity: <code>50</code>
-          </span>
-        </td>
-      </tr>
```
amyleadem, line 122: > [!Note]
> I removed these settings examples because they do not exist in USWDS and I couldn't find an alternative theme setting that accepted these tokens. The goal here is to reduce confusion so that users didn't try to bring a non-existent setting into their config. Please flag if you feel that removing these examples is too opinionated or would cause more confusion!

pages/design-tokens/opacity.md
amyleadem, line 1: > [!Note]
> I removed some settings examples in this PR because they do not exist in USWDS and I couldn't find an alternative theme setting that accepted these tokens. I was hoping to reduce confusion so that users didn't try to bring a non-existent setting into their config. Please flag if you feel that removing these examples is too opinionated or would cause more confusion!

amyleadem, line 1: > [!important]
> The `u-opacity` mixin does not currently exist in USWDS, and it looks like it hasn't for some time. This is likely a bug and is documented in https://github.com/uswds/uswds/issues/5881. Should we remove the reference to it here?

pages/design-tokens/opacity.md
amyleadem, line 122: > [!Note]
> I removed some settings examples in this PR because they do not exist in USWDS and I couldn't find an alternative theme setting that accepted these tokens. I was hoping to reduce confusion so that users didn't try to bring a non-existent setting into their config. Please flag if you feel that removing these examples is too opinionated or would cause more confusion!

pages/design-tokens/flex.md
```diff
@@ -126,7 +130,7 @@ Your context and coding style determine how you access USWDS flex tokens in code
         </td>
         <td data-title="Example">
           <span>
-            $theme-nav-column-flex: <code>4</code>
+            $theme-megamenu-columns: <code>4</code>
```
amyleadem, line 133: Not sure if this is a suitable replacement, but was the only setting I could find that was pulled into `u-flex`. Please let me know if this feels wrong to include here and/or if you have a suitable replacement.

_data/changelogs/tokens-flex.yml
```diff
@@ -0,0 +1,10 @@
+title: Flex tokens
+type: token
+changelogURL:
+items:
+  - date: NNNN-NN-NN
```
amyleadem, line 5: > [!important]
> We need to update the changelog dates before merge

amyleadem, line 5: >[!important]
> We need to update changelog dates before merge

mejiaj (approved):
Thanks for these fixes @amyleadem! Updated branch and approved.

PR #2396, opened 2023-12-12
USWDS-Site - Footer changelog: Update to include layout grid

mahoneycm:
# Summary

Updated changelog entry for uswds/uswds#5289 to include a note about layout grid.

## Preview link

[Footer changelog →](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/cm-changelog-footer-grid/components/footer/#latest-updates)

## Problem statement

Users who import footer but not `layout-grid` may experience visual regressions if they've:
- Modified standard breakpoint utility classes in the markup of the footer
- Use `layout-grid` utilities elsewhere on the site

## Solution

Update changelog to alert users of this change:

> **This update reduces the footer package size**. Thanks @danbrady! (#5289) <br><br> If you notice changes in your layout after making this update, your Sass might be missing the `usa-layout-grid` package. You can include it by adding `@forward usa-layout-grid` to your [Sass entry point](https://designsystem.digital.gov/documentation/migration/#using-component-packages).

## Testing and review

1. Inspect updated changelog
2. Confirm it is appearing appropriately
3. Approve spelling and grammar

### DIFF
--- a/_data/changelogs/component-footer.yml
+++ b/_data/changelogs/component-footer.yml
@@ -4,7 +4,7 @@ changelogURL:
 items:
   - date: 2023-12-08
     summary: Removed the `usa-layout-grid` dependency.
-    summaryAdditional: This update reduces the footer package size. If you notice changes in your layout after making this update, your Sass might be missing the usa-layout-grid package. You can include it by adding `@forward usa-layout-grid` to your [Sass entry point](https://designsystem.digital.gov/documentation/migration/#using-component-packages).
+    summaryAdditional: This update reduces the footer package size. If you notice changes in your layout after making this update, add the `usa-layout-grid` package to your [Sass entry point](https://designsystem.digital.gov/documentation/migration/#using-component-packages).
     affectsStyles: true
     githubPr: 5289
     githubRepo: uswds
--- a/_data/changelogs/component-footer.yml
+++ b/_data/changelogs/component-footer.yml
@@ -4,7 +4,7 @@ changelogURL:
 items:
   - date: 2023-12-08
     summary: Removed the `usa-layout-grid` dependency.
-    summaryAdditional: This update reduces the footer package size.
+    summaryAdditional: This update reduces the footer package size. If you notice changes in your layout after making this update, your Sass might be missing the usa-layout-grid package. You can include it by adding `@forward usa-layout-grid` to your [Sass entry point](https://designsystem.digital.gov/documentation/migration/#using-component-packages).
     affectsStyles: true
     githubPr: 5289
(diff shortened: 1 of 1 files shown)


_data/changelogs/component-footer.yml
```diff
@@ -4,7 +4,7 @@ changelogURL:
 items:
   - date: 2023-12-08
     summary: Removed the `usa-layout-grid` dependency.
-    summaryAdditional: This update reduces the footer package size.
+    summaryAdditional: This update reduces the footer package size. If you notice changes in your layout after making this update, your Sass might be missing the usa-layout-grid package. You can include it by adding `@forward usa-layout-grid` to your [Sass entry point](https://designsystem.digital.gov/documentation/migration/#using-component-packages).
```
amyleadem, line 7: This feels a bit long. Wondering if we can remove the last sentence and instead use the PR description to convey all the info they need to fix the issue. I think it is a logical step to click on "More information" if they are having trouble. 

Curious what you think!

```suggestion
    summaryAdditional: This update reduces the footer package size. If you notice changes in your layout after making this update, add the `usa-layout-grid` package to your Sass. 
```

amyleadem (changes requested):
@mahoneycm Added a comment to possibly shorten the entry.

_data/changelogs/component-footer.yml
mahoneycm: I'm also in favor of shortening the changelog. One concern I have is that if they are having trouble and click "More information", it brings them to the PR which doesn't provide additional troubleshooting. They might think "add the `usa-layout-grid` package to your Sass" might mean to add it it the _footer Sass_ since that's what is removed in the PR.

What do you think of keeping the link to the Sass entry point?

```suggestion
    summaryAdditional: This update reduces the footer package size. If you notice changes in your layout after making this update, add the `usa-layout-grid` package to your [Sass entry point](https://designsystem.digital.gov/documentation/migration/#using-component-packages).
```

Another thought I had was to word it as "If you relied on `footer` to import `layout-grid` styles, add the package..." but I feel like that's less succinct than addressing layout changes.

What do you think?

amyleadem: I like that suggestion. Can you update the PR description so that it has the information needed to fix the issue? Also, if you have an open issue for creating a fix for this, it would be good to point to it in the PR description as well.

amyleadem (approved):
Looks good to me!

mejiaj (approved):

PR #1362, opened 2021-12-06
Add Getting Started for Devs

thisisdano:
- Adds Getting Started content for developers
- Incorporates Ruby work from https://github.com/uswds/uswds-site/pull/1318

Replaces https://github.com/uswds/uswds-site/pull/1318 and https://github.com/uswds/uswds-site/pull/1360

[Preview](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.app.cloud.gov/preview/uswds/uswds-site/dw-add-getting-started-for-devs/documentation/getting-started-for-developers/) →

PR #1318, opened 2021-11-16
Bump ruby to 3.0.2

thisisdano:
This possibly addresses an local installation issue I was having with `liquid-c` similar to https://github.com/Shopify/liquid-c/issues/164

PR #2448, opened 2024-01-22
USWDS-Site - Security: Update load method to address security concerns.

mahoneycm:
# Summary

Adjusts ruby `open` methods to address [security concerns](https://github.com/uswds/uswds-site/security/code-scanning) flagged by CodeQL.

## Related issue

Addresses [5 security concerns](https://github.com/uswds/uswds-site/security/code-scanning) 🔒 detected by CodeQL

## Preview link

[Preview link →](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/cm-ruby-load-security-fix/)

## Problem statement

The method used to pull in some JSON data introduces vulnerabilities to the repo.

> Use of `Kernel.open` or `IO.read` or similar sinks with a non-constant value
> 

## Solution

Assess whether each `open` method is pulling from a local file or an HTTP request.

- When it’s a local json file, use `File.open(path)`

- When it’s from an HTTP request, use `URI(url).open`

### Additional Context

This change is based on the recommendation from CodeQL:

> **Recommendation**
Use File.open instead of Kernel.open, as the former does not have this vulnerability. Similarly, use the methods from the File class instead of the IO class e.g. File.read instead of IO.read.
>
> Instead of URI.open use URI(..).open or an HTTP Client.

We had originally changed all `open` methods to `URI.open(url)` in #1362. Based on [this comment](https://github.com/uswds/uswds-site/pull/1318/files#r750794675) it was to match a new Ruby syntax. When I updated the syntax to `URI(url).open` as recommended, some of the queries would return empty values.

Specifically, the queries that would cause the issue were searching the local `/jekyll_get_cache` or `html-templates` directories for JSON/HTML files respectfully. This was due to the query returning a [Same-Document Reference](https://www.rfc-editor.org/rfc/rfc3986#section-4.4) which would return empty with the new context, where previously it would return a hash value.

|`File.open(url)` | `URI.open(url)` | `URI(url).open` |
|--------|--------|--------|
| #<File:0x00000001052edfa0> | #<File:0x0000000106d874d8> | `empty` |

To avoid this, I replaced local queries with `File.open` and updated pulls from the Github repo with the `URI(url).open` syntax. This solution follows the recommendations provided by CodeQL without breaking the build process at any point.

## Major changes

- Code is improved to use variable names instead of chaining methods.
- Get cache functions are now using different open methods.

## Testing and review

> [!TIP]
You can check the url/var values by adding `puts *variable_name*` to the ruby code and running `bundle exec jekyll build`

This will run the jekyll build and print the variable value to the console as it’s building.
> 

- [ ]  Confirm `File.load(url)` is used only when we are loading local files.
- [ ]  Confirm `URI(url).load` is used only when we are loading data from the internet.
- [ ]  Rebuild cache
    - [ ]  Remove jekyll_get_cache using `rm -rf .jekyll_get_cache`
    - [ ]  Run `bundle exec jekyll build`
    - [ ]  Ensure component markup has compiled appropriately.
    - [ ]  Check `.jekyll_get_cache/` for the following json files
        - [ ]  `contributing.json`
        - [ ]  `hash.json`
        - [ ]  `install_readme.json`
        - [ ]  `releases.json`
        - [ ]  `security.json`


mejiaj (approved):
LGTM, I've run a clean install and no issues locally. Also tested Cloud Pages preview and no issues found.

**Pages tested**
Tested both locally and cloud pages.

- `documentation/developers/`
- `about/security/`
- `documentation/code-guidelines/`

These are the pages I've found to use the plugin to fetch data from GitHub.

thisisdano (approved):

thisisdano:
Good fix!

PR #2482, opened 2024-02-09
USWDS-Site - Changelog: Add entry for indeterminate checkboxes

mahoneycm:
# Summary

Adds changelog entry for uswds/uswds#5713

>  **Added indeterminate styles for checkboxes.** Checkboxes will now display as indeterminate when `input.indeterminate = true` is set via JS or the `data-indeterminate` attribute is present. This is currently only a style addition and does not affect checkbox functionality.

## Preview link

[Checkbox changelog →](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/cm-changelog-5713/components/checkbox/#latest-updates)


_data/changelogs/component-checkbox.yml
```diff
@@ -2,6 +2,14 @@ title: Checkbox
 type: component
 changelogURL:
 items:
+  - date: NNNN-NN-NN
+    summary: Added indeterminate styles for checkboxes.
+    summaryAdditional: Checkboxes will now display as indeterminate when `input.indeterminate = true` is set via JS or the `data-indeterminate` attribute is present. This is currently only a style addition and does not affect checkbox functionality.
```
amyleadem, line 7: Recommend we expand the JS abbreviation and move away from passive voice. 
```suggestion
    summaryAdditional: Checkboxes will now display as indeterminate when you set `input.indeterminate = true` via JavaScript or add the `data-indeterminate` attribute. This is currently only a style addition and does not affect checkbox functionality.
```

mahoneycm: Sounds good! Thanks for the tip

_data/changelogs/component-checkbox.yml
```diff
@@ -2,6 +2,14 @@ title: Checkbox
 type: component
 changelogURL:
 items:
+  - date: NNNN-NN-NN
+    summary: Added indeterminate styles for checkboxes.
+    summaryAdditional: Checkboxes will now display as indeterminate when you set `input.indeterminate = true` via JavaScript or add the `data-indeterminate` attribute. This is currently only a style addition and does not affect checkbox functionality.
```
amyleadem, line 7: Optional: One more question - wondering if we should announce the `data-attribute` here. Is this something that is good practice to use, or should users stick to the native `:indeterminate` state?
```suggestion
    summaryAdditional: Checkboxes will now display as indeterminate when you set `input.indeterminate = true` via JavaScript. This is currently only a style addition and does not affect checkbox functionality.
```

mahoneycm: That's a good point. Is there anything we should consider when setting `indeterminate` status via JS? You set the CSS pseudo-class via `input.indeterminate = true` 

Is there a JS benefit to allowing the use of a data attribute as well? 

If we deem that there isn't, should we remove support for the data attribute styles as well?

mejiaj: Initially, this seems fine to me. You can add the data attribute to get the initial styles first and then override when you need.

Something to keep in mind moving forward. Also totally possible that we revisit in the future once we have more information.

mejiaj (approved):

amyleadem (approved):
LGTM!

_data/changelogs/component-checkbox.yml
```diff
@@ -2,6 +2,14 @@ title: Checkbox
 type: component
 changelogURL:
 items:
+  - date: 2024-03-11
```
amyleadem, line 5: > [!important]
> We should update the changelog date before merge

thisisdano, line 5:
```suggestion
  - date: 2024-04-26
```

thisisdano (approved):

PR #2586, opened 2024-03-29
Remove broken links from Showcase page and correct spelling

finekatie:
# Summary

Remove broken links that have no alternate for redirects. Correct misspelling that Siteimprove noted. Will improve QA on site.

## Related issue #2567 
Also related to: #2346 and #2407

Closes #2567

## Preview link

[Showcase page](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/kf-showcase-linkremoval/documentation/showcase/)

## Problem statement

Broken links and misspelling on this page contributes to drop in QA.

## Solution

Removing old broken links and correcting spelling will improve content relevancy and QA.

## Changes

https://datacenters.cio.gov/	remove 
https://feedback.usa.gov/	remove
https://itmodernization.cio.gov/	remove 
https://www.move.mil/	remove
http://vsfs.state.gov/	remove

## Testing and review

Will need to have someone review to make sure links and copy are gone and that spelling is correct. 

Check that changelog is implemented correctly and the content is correct.

PR #2569, opened 2024-03-27
USWDS-Site: Remove outdated pages from search indexing

amyleadem:
# Summary
- **Added the `expiration_date` front matter data key.** The `expiration_date` key now controls when a page will be removed from search indexing. The date should be added in `YYYY-MM-DD` format.
- **Added expiration dates for posts from 2017-2021 in the ["News and updates" section of the site](https://designsystem.digital.gov/about/updates/).** These posts have an expiration date set in the past.

> [!Important]
> This update does not affect the presentation of the page. It only affects the meta data that tells whether search engines should crawl the page. 

## Related issue

Closes #2564

## Preview link
- [Post 1.0 Research Strategy Workshop](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-noindex/whats-new/updates/2017/03/01/post-1-0-research-strategy-workshop/)
- [NASA’s journey with the Standards](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-noindex/whats-new/updates/2017/03/21/nasa-journey-with-us-web-design-standards/)
- [Updating Design Assets](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-noindex/whats-new/updates/2017/03/21/update-design-assets/)
- [Q1 2017 Analytics Update](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-noindex/whats-new/updates/2017/03/27/u-s-web-design-standards-exposure-analytics/)
- [U.S. Customs and Border Patrol Case Study](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-noindex/whats-new/updates/2017/03/31/u-s-customs-and-border-patrol-case-study/)
- [Introducing Fractal and Federalist](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-noindex/whats-new/updates/2017/04/26/fractal-federalist/)
[...]

Issue #2565, opened 2024-03-27
USWDS-Site: Create a component that notifies users of outdated content


amyleadem:
## Summary
Create a component that notifies readers of outdated content.

Here is an example of an implementation of something similar from [digital.gov](https://digital.gov/2015/12/31/digitalgovs-2015-year-in-review/):
![image](https://github.com/uswds/uswds-site/assets/93996430/f4232f55-8a29-4f2b-9903-a3112d732f60)

brunerae:
Next steps:

- [ ] Copy from content team
- [ ] Update Component

sarah-sch:
@katharine-carr and @amyleadem - [sharing some copy ideas here](https://docs.google.com/document/d/1Y1HWbjVGI_nK7nFpYYmBaZT9RT6gypmBtXlTY3uAerE/edit?usp=sharing). Would love input.

sarah-sch:
@amyleadem - see [our top choice for copy in this document](https://docs.google.com/document/d/1Y1HWbjVGI_nK7nFpYYmBaZT9RT6gypmBtXlTY3uAerE/edit?usp=sharing):

This page is in the process of being archived. It was published X years ago and may include broken links or outdated information.

cc @finekatie

amyleadem:
@sarah-sch Thanks! I'll work on incorporating that soon so you can get a preview

PR #2588, opened 2024-04-02
USWDS-Site - Package.json: Cleanup HTML Proofer script

### LINKED ISSUE #2169: USWDS-Site - Feature: Revalidate ignore rules in HTMLProofer
### Is your feature request related to a problem? Please describe.

Revisit HTML Proofer `--ignore-files` rule in `package.json` to ensure they're still valid.

https://github.com/uswds/uswds-site/blob/a1d57a9a579aa4b525b7c90f8b0e0ce871039c70/package.json#L44

### Describe the solution you'd like

Only ignoring files that are absolutely necessary.

### Describe alternatives you've considered

n/a

### Additional context

_Originally posted by @amyleadem in https://github.com/uswds/uswds-site/pull/2167#pullrequestreview-1524677771_

### Code of Conduct

- [X] I checked the [current issues](https://github.com/uswds/uswds-site/issues) for duplicate feature requests.

mahoneycm:
# Summary

Removed unnecessary HTML Proofer `--ignore-files` and flags. The proofer still runs without error.

## Breaking change

This is not a breaking change.

## Related issue

Closes #2169

## Related pull requests

[USWDS-Site - HTML Proofer: Broken links April '24 #2589](https://github.com/uswds/uswds-site/issues/2589)

## Preview link

[Preview link →](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/cm-update-htmlproofer-script/)

## Problem statement

Some of the ignored files and CLI flags are outdated and no longer necessary.

## Solution

- Remove flags that represent [default values](https://github.com/gjtorikian/html-proofer#configuration). These values are defined as defaults for htmlproofer and can be safely removed.
    - `--allow-hash-href=true`
    - `--ignore-empty-alt=true`
- Remove ignored files that are not causing errors.
    - `/2019/`
    - `/select-a-language/`
- Remove hash selector from OMB pdf link that was causing the proofer to error out.
- Minor formatting adjustments for `--ignore-files`. This pattern matches `--ignore-urls` and other package scripts.
    
    ```diff
    - \"//2017//,//2018//,//code-guidelines//\"
    + '/2017/,/2018/,/code-guidelines/'
    ```
    
## Major changes

Additional pages and files are now being checked for broken links

```diff
- Checking 1349 external links
- Checking 916 internal links
- Checking internal link hashes in 211 files
- Ran on 342 files!

+ Checking 1381 external links
+ Checking 920 internal links
+ Checking internal link hashes in 217 files
+ Ran on 348 files!
```

## Testing and review

1. Confirm running `npm run proof` results in 0 errors
3. Approve of updated `--ignore-files` formatting.
4. Confirm no CI failures

### Status code `0`

Removing `--ignore-status-codes 0` flags a few links as broken whcih are not. It also appropriately caught some broken links. Opting to keep ignoring the status code due to the false errors and updating the actual broken links.

I've created #2589 to capture this issue and track the work.

~### `select-a-language` errors~

~Removing `/select-a-language/` from the ignored files causes html proofer to crash due to a nokogiri issue with an HTML element having too many attributes. It appears that it was originally ignored in [this commit](https://github.com/uswds/uswds-site/commit/58470914893f77110a2e513be7133b2171424a06) due to this error.~

_Resolved as a part of this PR_

### Errors in `2017/2018` blog posts

All the proofing errors from the blogposts originate from outdated federalist links

- `federalist.18f.gov`
- `federalist.fr.cloud.gov`

Instead of ignoring the blog posts, we could just opt to ignore these broken links by adding them to `--ignore-urls`

~### `code-guidelines` error~

~There is only one error in the `code-guidelines` pages that is a false positive for the USWDS license [public domain hash](https://github.com/uswds/uswds/blob/develop/LICENSE.md#the-rest-of-this-project-is-in-the-worldwide-public-domain) that does work as expected. I was unable to determine the source of the error.~

_Resolved by [af84731](https://github.com/uswds/uswds-site/pull/2588/commits/af847317ebe16de1177b33e68355040b0de092e4)_


mejiaj (approved):
The explicit flags were nice because you can immediately understand what's being checked. The small downside being that it made the script pretty long. We can run with this shorter version now.

Also, I ran `npm run proof` and saw a _tiny_ difference in tests, but no issues there.


**Before/after**

```diff
- Checking 1349 external links
- Checking 916 internal links
- Checking internal link hashes in 211 files
- Ran on 342 files!

+ Checking 1357 external links
+ Checking 916 internal links
+ Checking internal link hashes in 212 files
+ Ran on 343 files!
```

mahoneycm:
@mejiaj That is correct! Additional files and pages are being checked for broken links as a result of removing some of the ignore rules. I added your diff block to the PR description for added visibility 👍

package.json
```diff
@@ -41,7 +41,7 @@
     "federalist": "npx gulp build",
     "lint": "npx gulp lintJS lintSass && npm run prettier:scss",
     "prestart": "npx gulp build",
-    "proof": "bundle exec htmlproofer --allow-hash-href=true --enforce-https=false --allow-missing-href=true --ignore-empty-alt=true --ignore-status-codes 0,403,429,503,302 --swap-urls 'https\\://designsystem.digital.gov/:/' --ignore-urls '/github.com/uswds/uswds/issues/' --ignore-files \"//select-a-language//,//2017//,//2018//,//2019//,//code-guidelines//\" ./_site",
+    "proof": "bundle exec htmlproofer --enforce-https=false --allow-missing-href=true --ignore-status-codes 0,403,429 --swap-urls 'https\\://designsystem.digital.gov/:/' --ignore-urls '/github.com/uswds/uswds/issues/' --ignore-files '/select-a-language/,/2017/,/2018/,/code-guidelines/' ./_site",
```
amyleadem, line 44: Hi @mahoneycm, I was hoping we could remove the `/select-a-language` directory from the "ignore files" config here as well so that we don't miss any more broken links on these pages. 

I did a little bit of poking around and found that the related nokogiri error goes away when I remove the `#page=16` hash on this link on the [Preferred language page](https://designsystem.digital.gov/patterns/select-a-language/language-preferences/#references): `https://www.whitehouse.gov/wp-content/uploads/legacy_drupal_files/omb/memoranda/2017/m-17-06.pdf#page=16`. Not quite sure why this removes the error, but curious if we could find a way to edit this link and remove that directory from the ignored files list. Curious what you think!

```suggestion
    "proof": "bundle exec htmlproofer --enforce-https=false --allow-missing-href=true --ignore-status-codes 0,403,429 --swap-urls 'https\\://designsystem.digital.gov/:/' --ignore-urls '/github.com/uswds/uswds/issues/' --ignore-files '/2017/,/2018/,/code-guidelines/' ./_site",
```

mahoneycm: Thank you SO much for finding this! It was a big thorn in my side. I'll do some additional poking and prodding. I'm assuming it's because this links a to a PDF and the page number is set using a hash but there is no associated heading level like there would be for a webpage

mahoneycm: @amyleadem ok so I opted to remove the page number. I feel like removing the page number is ideal so that we can continue to proof the link. 

I was able to successfully ignore just this link though if we want to continue to link to that specific page, we could instead update the script to ignore it.

I also tried alternative means of creating a jump link to that section but it doesn't look like we can without editing the actual PDF.

pages/patterns/select-language/language-preferences.md
```diff
@@ -134,7 +134,7 @@ The pattern allows users to select their language preferences for both written a
 - Improving access to public websites and digital services for limited English proficient (LEP) persons. (December 2021) Retrieved on July 25, 2022, from <https://www.lep.gov/sites/lep/files/media/document/2021-12/2021_12_07_Website_Language_Access_Guide_508.pdf>
 - LEP.gov. (n.d.) Retrieved on July 25, 2022, from <https://www.lep.gov/>
 - The limited English proficient population in the United States in 2013. (July 8, 2015) Retrieved on July 25, 2022, from <https://www.migrationpolicy.org/article/limited-english-proficient-population-united-states-2013>
-- OMB memorandum m-17-06, 16. (November 2016) Retrieved on July 25, 2022, from <https://www.whitehouse.gov/wp-content/uploads/legacy_drupal_files/omb/memoranda/2017/m-17-06.pdf#page=16>
+- OMB memorandum m-17-06, 16. (November 2016) Retrieved on July 25, 2022, from <https://www.whitehouse.gov/wp-content/uploads/legacy_drupal_files/omb/memoranda/2017/m-17-06.pdf>
```
mahoneycm, line 137: Should we keep the page number in the citation? It might be helpful to maintain that reference since we ran into errors linking directly to the page.

Alternatively, we can remove the page number from the reference.

```suggestion
- OMB memorandum m-17-06. (November 2016) Retrieved on July 25, 2022, from <https://www.whitehouse.gov/wp-content/uploads/legacy_drupal_files/omb/memoranda/2017/m-17-06.pdf>
```

amyleadem: @sarah-sch @finekatie We had a couple of questions about this link. A little background: We have had a long-standing error that prevents us from running our automated HTML proofer check on the "select-a-language" directory on our site, related to the "#page=16" originally added to this link  (See [this comment](https://github.com/uswds/uswds-site/pull/2588#discussion_r1563077334) for more detail). We were wondering - is it ok to remove the page reference from the link and just point to the pdf main page? And if so, should we change the link text? I'm wondering if it might help to add the page number to the link text, maybe something like this:

```suggestion
- OMB memorandum m-17-06,  page 16. (November 2016) Retrieved on July 25, 2022, from <https://www.whitehouse.gov/wp-content/uploads/legacy_drupal_files/omb/memoranda/2017/m-17-06.pdf>
```

Curious what you think!

mahoneycm: Just adding a note that this is the lead text, not the link text itself

image

finekatie: > @sarah-sch @finekatie We had a couple of questions about this link. A little background: We have had a long-standing error that prevents us from running our automated HTML proofer check on the "select-a-language" directory on our site, related to the "#page=16" originally added to this link (See [this comment](https://github.com/uswds/uswds-site/pull/2588#discussion_r1563077334) for more detail). We were wondering - is it ok to remove the page reference from the link and just point to the pdf main page? And if so, should we change the link text? I'm wondering if it might help to add the page number to the link text, maybe something like this:
> 
> Curious what you think!

I think it works fine to take it out of the link and have "page 16" after the OMB title. Just "16" by itself would be a little hard to tell, but Amy's change works for me.

package.json
```diff
@@ -41,7 +41,7 @@
     "federalist": "npx gulp build",
     "lint": "npx gulp lintJS lintSass && npm run prettier:scss",
     "prestart": "npx gulp build",
-    "proof": "bundle exec htmlproofer --allow-hash-href=true --enforce-https=false --allow-missing-href=true --ignore-empty-alt=true --ignore-status-codes 0,403,429,503,302 --swap-urls 'https\\://designsystem.digital.gov/:/' --ignore-urls '/github.com/uswds/uswds/issues/' --ignore-files \"//select-a-language//,//2017//,//2018//,//2019//,//code-guidelines//\" ./_site",
```
amyleadem, line 44: > [!Note]
> I think that removing the 2019 directory is ok for now, but once we remove older pages from search indexing (https://github.com/uswds/uswds-site/pull/2569) and add an alert about outdated content (#2565), I vote we restore the 2019-2021 directories in the ignored files. We can have that discussion as part of #2569.

package.json
```diff
@@ -41,7 +41,7 @@
...
+    "proof": "bundle exec htmlproofer --enforce-https=false --allow-missing-href=true --ignore-status-codes 0,403,429 --swap-urls 'https\\://designsystem.digital.gov/:/' --ignore-urls '/github.com/uswds/uswds/issues/' --ignore-files '/2017/,/2018/,/code-guidelines/' ./_site",
```
amyleadem, line 44: When I remove `/code-guidelines` from the ignore list and run `npm run proof` locally, I get a single failure with status code 200. [Status code 200](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200) means "successful", and I wonder if it is safe to remove `code-guidelines` from the ignore list and instead add status 200.

``` 
* At ./_site/documentation/code-guidelines/index.html:2815:

  External link https://github.com/uswds/uswds/blob/develop/LICENSE.md#the-rest-of-this-project-is-in-the-worldwide-public-domain failed: https://github.com/uswds/uswds/blob/develop/LICENSE.md exists, but the hash 'the-rest-of-this-project-is-in-the-worldwide-public-domain' does not (status code 200)
```

```suggestion
    "proof": "bundle exec htmlproofer --enforce-https=false --allow-missing-href=true --ignore-status-codes 0,200,403,429 --swap-urls 'https\\://designsystem.digital.gov/:/' --ignore-urls '/github.com/uswds/uswds/issues/' --ignore-files '/2017/,/2018/' ./_site",
```

amyleadem (changes requested):
@mahoneycm I ran through all of the ignores in my local build and added some comments below. Most don't require action, but I wanted to flag.

package.json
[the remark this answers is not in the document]
mahoneycm: Thank you for the extra attention here! I had originally made this a part of the testing instructions to approve that this was indeed a change that brought potential value rather than a headache down the road.

> 2. Approve of the removal of error codes 503, 302. Would we want to keep these codes in the event one of the errors pop up in the future?
>    - 503 - This code indicates the server is not ready to handle the request
>    - 302 - This code indicates a specific URL has been moved temporarily to a new location
> 

Since these are temporary errors, how do you feel about adding them back to the ignore script?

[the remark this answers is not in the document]
mahoneycm: I created #2589 to track this work when I created this PR and left a note in the **Additional Findings** section. These links will be resolved when these PRs are merged:

- [ ] #2586 
- [ ] #2609 

I agree that we should continue to ignore this

> Opting to keep ignoring the status code due to the false errors and updating the actual broken links.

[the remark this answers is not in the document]
mahoneycm: Since we both see value in keeping them ignored, I went ahead and restored 👍

package.json
mahoneycm: Resolved by [ba50af9](https://github.com/uswds/uswds-site/pull/2588/commits/ba50af9fcde967b2d16aaf2b609d3ce4688bcbd1)

package.json
```diff
@@ -41,7 +41,7 @@
     "federalist": "npx gulp build",
     "lint": "npx gulp lintJS lintSass && npm run prettier:scss",
     "prestart": "npx gulp build",
-    "proof": "bundle exec htmlproofer --allow-hash-href=true --enforce-https=false --allow-missing-href=true --ignore-empty-alt=true --ignore-status-codes 0,403,429,503,302 --swap-urls 'https\\://designsystem.digital.gov/:/' --ignore-urls '/github.com/uswds/uswds/issues/' --ignore-files \"//select-a-language//,//2017//,//2018//,//2019//,//code-guidelines//\" ./_site",
+    "proof": "bundle exec htmlproofer --enforce-https=false --allow-missing-href=true --ignore-status-codes 0,403,429,503,302 --swap-urls 'https\\://designsystem.digital.gov/:/' --ignore-urls '/github.com/uswds/uswds/issues/' --ignore-files '/2017/,/2018/,/code-guidelines/' ./_site",
```
mahoneycm, line 44: ## Status code 200 and code-guidelines

Opted to maintain `code-guidelines` rather than ignoring code `200` to ensure we catch incorrect hashes on external links

amyleadem (approved):
Looks good to me! 

We'll need to resolve the snyk errors in a separate PR (I'll work on that shortly). 

I'd also like to see @mejiaj re-review after the recent changes.

package.json
mejiaj: @mahoneycm shouldn't this be `/documentation/code-guidelines/`? The current url returns a 404.

pages/patterns/select-language/language-preferences.md
```diff
@@ -134,7 +134,7 @@ The pattern allows users to select their language preferences for both written a
 - Improving access to public websites and digital services for limited English proficient (LEP) persons. (December 2021) Retrieved on July 25, 2022, from <https://www.lep.gov/sites/lep/files/media/document/2021-12/2021_12_07_Website_Language_Access_Guide_508.pdf>
 - LEP.gov. (n.d.) Retrieved on July 25, 2022, from <https://www.lep.gov/>
 - The limited English proficient population in the United States in 2013. (July 8, 2015) Retrieved on July 25, 2022, from <https://www.migrationpolicy.org/article/limited-english-proficient-population-united-states-2013>
-- OMB memorandum m-17-06, 16. (November 2016) Retrieved on July 25, 2022, from <https://www.whitehouse.gov/wp-content/uploads/legacy_drupal_files/omb/memoranda/2017/m-17-06.pdf#page=16>
```
mejiaj, line 137: Shame we had to get rid of this, but it seems to be causing an error:

```bash
htmlproofer 4.4.3 | Error:  Attributes per element limit exceeded
/Users/jmejia-a/.rvm/gems/ruby-3.1.2/gems/nokogiri-1.15.4-arm64-darwin/lib/nokogiri/html5/document.rb:123:in `parse': Attributes per element limit exceeded (ArgumentError)
```

Not seeing any related open issues of it in HTMLProofer or Nokogiri.

mejiaj (changes requested):
Looks good, no issues found testing locally. Minor change on code guidelines.

package.json
mahoneycm: @mejiaj Since we're specifying file paths and not urls here, having `code-guidelines` is enough to properly target the correct page. Testing both and comparing the number of files ran as well as the test results shows no difference with either option.

Would you prefer the added specificity of `/documentation/code-guidelines/` ?

mejiaj: Thanks for that note. I wasn't aware about file paths. Yes, I think so since it's the more accurate option.

FWIW on site build I still see this structure:

```bash
.
└── _site
    └── documentation
        └── code-guidelines
            └── index.html
```


image

mahoneycm: Resolved in [9828b79](https://github.com/uswds/uswds-site/pull/2588/commits/9828b798f554523a3c80f243f4059e56d95f6704). 

I gave the `2017, 2018` links the same treatment for consistency 👍

mejiaj (approved):
Looks good, thank you.

thisisdano:
Merged in lated `main` — this should resolve our pa11y test fails

thisisdano (approved):

Issue #2543, opened 2024-03-20
Add links to sub-pages on Performance landing page


brunerae:
The [Performance page](https://designsystem.digital.gov/performance/) needs to be updated to include links to its sub-pages.

ryparker:
Consolidating into #3195. Both issues request links from the Performance landing page to its sub-pages. #3195 is the current tracking issue and includes the relationship to #2542. This is consolidation, not a claim that those links are implemented.

PR #2595, opened 2024-04-02
USWDS-Site: Restore in-page navigation on long pages

### LINKED ISSUE #2594: USWDS-Site - Enhancement: Restore in-page nav to long pages
## Summary
We removed the in-page navigation component from pages when the header list was too long and extended beyond the viewport. We should use the new `data-heading-elements` attribute on the in-page navigation component to allow us to turn the component on for these long pages.

## Tasks
- [ ] Identify which pages should have in-page nav turned on 
- [ ] Create a way to designate which headers should be included in the component with `data-heading-elements`

amyleadem:
# Summary
- **Added an option in front matter to designate which headers should be included in the page's in-page navigation component's link list.**
- **Updated the `in_page_nav` front matter variable name to `in_page_nav_headings`.** This will improve clarity about the intent of the variable. 
- **Restored the in-page navigation component to pages with long header lists.**
- **Updated the date on the a11y page changelog.**

## New value options
The `in_page_nav` front matter data key now accepts header levels in its value definition. If the value is anything other than `false`, it will add that value to the in-page navigation's `data-heading-elements` attribute, as [documented on our website](https://designsystem.digital.gov/components/in-page-navigation/#in-page-navigation-initialization-properties). If `in_page_nav` is not defined, the in-page navigation component will automatically be added to the page and will show both `h2` and `h3` headers in the link list. 

Refer to the table below for expected implementations of `in_page_nav`:

| Implementation | Meaning |
| -------- | ------- |
| in_page_nav_headings: false | Remove in-page navigation component from the page. This will still be used for pages that have no headers. |
| in_page_nav_headings: "h2" | Show only `h2` headers in the page's in-page navigation link list. This should be used when the page's heading structure is very long and causes the link list to extend beyond the viewport. |
| in_page_nav_headings: "h3" | Show only `h3` headers in the page's in-page navigation link list. This use case will be rare. |

## Related issue

Closes #2594

## Preview link

- [Accessibility page](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-in-page-nav-headers/documentation/accessibility/)
- [Header component page](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-in-page-nav-headers/components/header/#basic-header-with-megamenu-2)
- [Language selector component page](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-in-page-nav-headers/components/language-selector/)
- [Migrating to USWDS 2.0](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-in-page-nav-headers/documentation/migration-v2/)
- [Settings page](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-in-page-nav-headers/documentation/settings/)

## Problem statement

Previously, we removed the in-page navigation component from pages with a long list of headers that caused the component to beyond the viewport. We should use the new `data-heading-elements` attribute on the in-page navigation component to show only `h2` elements in the link list. 
## Solution

Adding the option to show only `h2` headers on the in-page navigation component instead of the default `h2` and `h3` headers allows us to shorten the in-page navigation link list and display the component on the page.

## Testing and review

- Confirm that the in-page navigation component shows up on the pages listed in "Preview links" above
- Confirm that the pages that need an in-page navigation component have one
- Confirm that there are no regressions with the in-page nav on other pages

### Devs 
- Confirm that the code added to [styleguide.html](https://github.com/uswds/uswds-site/pull/2595/files#diff-08def7b13077a289e2dcb786dd5cd1ecf5cf2f821652a0a2b75555034ef463f3) meets code standards


_layouts/styleguide.html
```diff
@@ -17,7 +17,9 @@
       aria-label="On this page"
       data-scroll-offset="-120"
       data-root-margin="48px 0px -90% 0px"
-      data-threshold="1"></aside>
+      data-threshold="1"
+      data-heading-elements={{ page.in_page_nav }}
```
amyleadem, line 21: I considered adding a check here for certain values, but ultimately moved away from this because it felt unnecessarily constraining. Let me know if I should add something like this in!
```suggestion
      {% if page.in_page_nav == "h2" or page.in_page_nav == "h3" %}
          data-heading-elements={{ page.in_page_nav }}
      {% endif %}
```

mahoneycm (approved):
This is looking good! While testing I noticed a few visual discrepancies that might be out of scope for this issue but I wanted to document

## Testing checklist
- [x] Confirm that the in-page navigation component shows up on the pages listed in "Preview links" above
- [x] Confirm that the pages that need an in-page navigation component have one
- [x] Confirm that there are no regressions with the in-page nav on other pages
- [x] Confirm that the code added to [styleguide.html](https://github.com/uswds/uswds-site/pull/2595/files#diff-08def7b13077a289e2dcb786dd5cd1ecf5cf2f821652a0a2b75555034ef463f3) meets code standards

## Additional findings

### Deactivate in-page nav on [What is web performance](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-in-page-nav-headers/performance/what/)

I noticed this page has an empty in-page navigation that should be removed. It's a small change but we could also add it to #2543 or create another issue if it's out of scope for this ticket.

### Add setting to select breakpoint that activates in-page nav

This is a larger lift that should be documented in a new issue if we choose to implement it. Currently, the in-page nav component is set to activate at `tablet` widths. I noticed the display of the [component overview page](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-in-page-nav-headers/components/overview/) starts to break just over the `tablet` viewport.

image

What do you think about adding an additional attribute or theme setting to update the breakpoints for in-page navs?

pages/ui-components/overview.md
```diff
@@ -8,7 +8,7 @@ type: docs
 redirect_from:
   - /components/
   - /form-controls/
-in_page_nav: false
+in_page_nav: "h2"
```
sarah-sch, line 11: I don't think we need a side nav on this page. The invitation to "Find a USWDS Component" already appears "above the fold," so IMO the sidenav is redundant and looks little awkward.

_templates/page-templates/overview.md
```diff
@@ -7,7 +7,6 @@ title: Templates
 category: Templates
 lead: These templates represent common component combinations and are a great starting point for prototyping or trying USWDS to see if it’s a good fit for your project. Every template in this section is built using USWDS default theme settings and utilities.
 type: docs
-in_page_nav: false
```
sarah-sch, line 10: It looks like the in-page nav may have already been removed from the templates overview page, but I would note that I don't think this page is long enough to warrant in-page nav.

_components/header/header.html
```diff
@@ -32,7 +32,7 @@
     description: Display an extended header.
 changelog:
   key: component-header
-in_page_nav: false
```
sarah-sch, line 35: Thanks for adding the in-page nav here. I see some issues with the heading levels here, and I'm not sure this PR is the place to fix them, but I wanted to call them out. I think the "Component preview," "Component code," "Guidance," etc., headings should be nested under each heading type (e.g., nested under "Basic header"). To me, it is not logical for these to be at the same level as the heading type--they belong under the heading type.

_components/language-selector/language-selector.md
```diff
@@ -40,7 +40,7 @@ title: Language selector
 type: component
 changelog:
   key: component-language-selector
-in_page_nav: false
```
sarah-sch, line 43: Similar to the header nav, I don't think the preview, code, guidance headers, etc. should have the same heading level as the language selector type under which they fall. This guidance should be nested at a lower heading level.

sarah-sch (changes requested):
Requesting some changes. Some of these may be better handled in a new issue, so just let me know if I can help write one. I do not have any additional pages to add--you've covered the ones I had noted as needing in-page-nav in the audit. Thank you!

pages/ui-components/overview.md
amyleadem: Removed in [cd074b6](https://github.com/uswds/uswds-site/pull/2595/commits/cd074b67d0a0f0cc64d8ac42066ca8b06c50793e)

_templates/page-templates/overview.md
amyleadem: Removed in [cd074b6](https://github.com/uswds/uswds-site/pull/2595/commits/cd074b67d0a0f0cc64d8ac42066ca8b06c50793e)

_components/header/header.html
amyleadem: @sarah-sch I opened issue #2601 to update the header structure for these scenarios.

amyleadem:
@sarah-sch I have made the requested updates and opened a new issue (#2601) to address the component page's header structure. This is ready for your re-review.

sarah-sch (approved):
Thanks, @amyleadem! As long as we can work towards improving the nesting of headings on those longer component pages I think these are good to publish.

_components/header/header.html
```diff
@@ -32,7 +32,7 @@
...
+in_page_nav: "h2"
```
mejiaj, line 35: I'd like to see a little more specificity in the variable name that's used. Currently, it's not immediately clear what the `in_page_nav` is supposed to do. Previously, it was a boolean, but now it accepts strings?

```suggestion
in_page_nav_headings: "h2"
```

Other suggestions:
1. `in_page_nav_targets`
2. `in_page_nav_controls` 
3. `in_page_nav_capture`

_layouts/styleguide.html
mejiaj: What would this change do?

mejiaj (commented):
Looks good. 

I was able to confirm in-page nav is now displaying correctly on these pages. Minor  suggestion on variable name.

If you disagree with it, please comment and we can resolve.

### Pages tested

| Page              | Old                                                                                           | New                                                                                                                                                                                                                      |
| :---------------- | :-------------------------------------------------------------------------------------------- | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Accessibility     | [Live →](https://designsystem.digital.gov/documentation/accessibility/)                    | [Preview →](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-in-page-nav-headers/documentation/accessibility/#we-follow-the-four-principles-of-accessibility-2) |
| Header            | [Live →](https://designsystem.digital.gov/components/header/#basic-header-with-megamenu-2) | [Preview →](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-in-page-nav-headers/components/header/#basic-header-with-megamenu-2)                               |
| Language selector | [Live →](https://designsystem.digital.gov/components/language-selector/)                   | [Preview →](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-in-page-nav-headers/components/language-selector/)                                                 |
| Migrating         | [Live →](https://designsystem.digital.gov/documentation/migration-v2/)                     | [Preview →](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-in-page-nav-headers/documentation/migration-v2/)                                                   |
| Settings          | [Live →](https://designsystem.digital.gov/documentation/settings/)                         | [Preview →](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-in-page-nav-headers/documentation/settings/)                                                       |

_components/header/header.html
amyleadem: Good point. I updated the name to `in_page_nav_headings` in [c3b1952](https://github.com/uswds/uswds-site/pull/2595/commits/c3b195241a5bb87dd49e12951847be491039de2f).

Note that the variable still accepts a value of `false`, which will remove it from the page. Is there any issue with this var accepting both strings and boolean values?

_layouts/styleguide.html
amyleadem: The update shown in the suggested change would just prevent the attribute from being added unless the variable receives one of the expected values.

The benefit of this is that the rendered code would be cleaner (the attribute wouldn't be added with an empty value). 

The downside of this is that it either limits the possible values we can enter OR we'd have to check for all possible combinations here. For example, maybe we will have a need to show `h3` and `h4` values in the in-page nav on a given page, or maybe we would want `h2`, `h3`, and `h4`. 

I opted for an open value for the attribute here for maximum flexibility.

_components/header/header.html
mejiaj: Sounds good with the added clarity in variable name. We can follow-up in the future if this becomes an issue.

Thanks!

mejiaj (approved):
> [!NOTE]  
> There's an HTMLProofer error resolved by #2604.

Thanks for updating, approved.

PR #2621, opened 2024-04-17
USWDS-Site - Content: Add note about incorrect package date to download page

### LINKED ISSUE #2499: USWDS-Site - Documentation: Add note about incorrect package dates
# Summary

Add note about incorrect package dates to download page. Note provided by content strategy:

> Note: The file date defaults to October 26, 1985, when you download and install this package. This is intentional. NPM sets a default date to ensure reproducible builds for npm packages.

[Download and Documentation page updates 2-22-2024](https://docs.google.com/document/d/1hOHNJ8j8knySKZ-C6mxFaMxalLC5q3-_VkqbHIK8IMg/edit) (🔒)

## Problem statement

NPM purposefully sets package dates to 10/26/1985 to ensure hash-identical tarballs no matter when or on what device the `npm pack` is ran. This can confuse users who inspect the package and see the incorrect files for the date.

## Proposed Solution

Add note to site to specify this is intended behavior from NPM.

## Affected pages

[Download page →](https://designsystem.digital.gov/download/)
[Documentation page →](https://designsystem.digital.gov/documentation/developers/#install-the-package-directly-from-github-2)

## Related issue

Since the documentation page content is pulled from the [USWDS README](https://github.com/search?q=repo:uswds/uswds%20readme&type=code), we'll need to update the USWDS repo as well.

- [ ] uswds/uswds#5872

mahoneycm:
# Summary
Added site note to download page addressing incorrect package dates.

> [!IMPORTANT]
We are adding another note to the Documentation page which is generated by USWDS README content. This work is being addressed in uswds/uswds#5871
> 

## Related issue

Closes #2499

## Related pull requests

uswds/uswds#5871

## Preview link

[Preview link →](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/cm-note-incorrect-download-dates/download/)

## Problem statement

NPM purposefully sets package dates to 10/26/1985 to ensure hash-identical tarballs no matter when or on what device the `npm pack` is ran. This can confuse users who inspect the package and see the incorrect files for the date.

## Solution

Add note to site to specify this is intended behavior from NPM.

## Testing and review

1. Visit download page
2. Confirm note is showing
3. Confirm there are no grammatical errors
4. Confirm note is styled appropriately
5. Approve of changelog file and display

> [!TIP]
There is a slight indentation on the version and hash numbers. If we want the note to match this style, I can add the `link-download-subtext` class, or similar margin utility class
>


mejiaj (changes requested):
Thanks for creating the note. I'd like to see if we can do a plain language pass.

Essentially, the date isn't what users are expecting and this is expected because of how the code is package and shipped.

pages/documentation/download.html
```diff
@@ -24,6 +24,8 @@ <h2 class="usa-heading">Code</h2>
 
 <!--<span class="link-download-subtext">Version {{ site.uswds_version }}</span>-->
 
+<div class="site-note"><strong>Note:</strong> The file date defaults to October 26, 1985, when you download and install this package. This is intentional. NPM sets a default date to ensure reproducible builds for npm packages.</div>
```
sarah-sch, line 27: This is the re-write Amy L. suggested. I think it's a good improvement:

Note: Files in the downloadable USWDS package will show a "last modified" date of October 26, 1985. This is intentional. Npm sets this default date on all its packages to ensure reproducible builds.

We should update the capitalization of npm at the beginning of sentences elsewhere on this page or rewrite to avoid putting npm at the beginning of sentences. Should we do that here or in another PR?

sarah-sch (changes requested):

pages/documentation/download.html
finekatie: Just dropping in my two cents with a small tweak to Amy's language:

Note: Files in the downloadable USWDS package will show a "last modified" date of October 26, 1985. This is intentional. This default date is set by npm on all its packages to ensure builds can be duplicated.

I moved "npm" in the second sentence so we wouldn't have to capitalize, and I modified how we talk about something being duplicated - "reproducible" is a tougher word to get through. If that's technically inaccurate, we can leave it. It still checks out in Readability app for good reading level.

finekatie (commented):
commented in thread

pages/documentation/download.html
mahoneycm: @finekatie what do you think about "ensure builds are identical" instead? From my understanding, I think that's more on the nose in this scenario 🤔

mahoneycm: Resolved in https://github.com/uswds/uswds-site/pull/2621/commits/a227cd7e061f2d768962946c7816f482b8e52219

amyleadem:
@mahoneycm Is the intention to have the same note in both the README and the download page? If so, I wonder if we can pull in the README content onto the download page so that we don't have to do dual maintenance if we ever need to update the copy. 

We should be able to isolate the note in the README by using `split`, like we do on the [documentation page](https://github.com/uswds/uswds-site/blob/main/pages/documentation/developers.md?plain=1#L34). 

Do you see any disadvantages to this approach?

sarah-sch (approved):
LGTM

mejiaj (changes requested):
Can we try Amy's idea in comment to re-use content and avoid having to maintain it in multiple places?

finekatie (approved):
LGTM. I don't know about the solution Amy proposed, but if it allows us to publish once and display in multiple places, that's great.

mahoneycm:
## Using liquid `split` to isolate note from USWDS README

After testing, it unfortunately doesn't seem possible to isolate the note form the USWDS readme without adding additional HTML or by adding a very specific `slice` feature to slice the note using the character count of the section.

In the [Documentation page](https://github.com/uswds/uswds-site/blob/main/pages/documentation/developers.md?plain=1#L34) we are able to split the content between two heading elements that are created via markup. Since there is no markup that we can use to exclusively target the note.

Further, even with the following HTML, it proved difficult to isolate the note markup without causing visual regressions. 

**Example:** 
```html
<div id="npm-note"> <!-- Used to target the note with `split`. Cannot use p class due to it being removed during split -->
  <p class="site-note">
    <strong>Note:</strong> Files in the downloadable USWDS package will show a "last modified" date of October 26, 1985. This is <a href="https://github.com/npm/npm/issues/20439#issuecomment-385121133">intentional</a>. This default date is set by npm on all its packages to ensure builds will be identical.
  </p>
</div>
```

> [!IMPORTANT]
> Since we were unable to maintain this information in one location, I added a comment to ensure these notes match across repos

amyleadem (approved):
Looks good to me! Confirmed it matches the text in https://github.com/uswds/uswds/pull/5871. Thanks for adding those comments.

pages/documentation/download.html
```diff
@@ -24,6 +24,12 @@ <h2 class="usa-heading">Code</h2>
 
 <!--<span class="link-download-subtext">Version {{ site.uswds_version }}</span>-->
 
+<!-- 
+  This note also exists on the Documentation page, which is populated by the USWDS README. 
+  If any changes are made, make sure to update there as well. 
+-->
```
mejiaj, line 30: Following [liquid docs](https://shopify.github.io/liquid/tags/template/), we can use a comment that won't render. 

```suggestion
{% comment %}
  This note also exists on the Documentation page, which is populated by the USWDS README.
  If any changes are made, make sure to update there as well.
{% endcomment %}
```

image

mejiaj (approved):
Approved with minor suggestion.

amyleadem:
@mahoneycm Can you create a changelog entry for this change?

amyleadem:
Also, just flagging that the base branch for this is `main`. I am not sure if we wanted it to be merged with other 3.8.1. changes in the `release-3.8.1` branch. Just flagging for consideration.

_data/changelogs/about-download.yml
mahoneycm, line 1: ## Changelog

I created a new changelog for the download page and this update! I opted to name this as `about-download` since it lives in the "How to use USWDS" section of the site.

_data/changelogs/about-download.yml
```diff
@@ -0,0 +1,9 @@
+title: Download
+type: documentation
+changelogURL:
+items:
+  - date: NNNN-NN-NN
```
mahoneycm, line 5: > [!IMPORTANT]
> Changelog date should be updated before merge

thisisdano, line 5:
```suggestion
  - date: 2024-05-15
```

_data/changelogs/about-download.yml
```diff
@@ -0,0 +1,9 @@
...
+    summary: Added note about incorrect package dates in downloadable USWDS package.
+    affectsGuidance: true
```
thisisdano, line 7:
```suggestion
```

Lets only include this one when there is a meaningful guidance change (ie something that would change how we suggest teams do something, like use a component). I think this copy clarification doesn't pass that test.

thisisdano (approved):

thisisdano:
Merging over pa11y crashes

PR #2699, opened 2024-05-29
Improve Pa11y and Circle reliability

thisisdano:
I modified a Puppeteer Chrome config in our `.pallyci` configs based on an example I found at  https://opensource.com/article/23/2/automated-accessibility-testing and it seemed to make a difference in the reliability of Pa11y. I also boosted the concurrency and used `npm ci` instead of `npm install` so the `package-lock` checksum–based cache actually works.

It seems the real solution here is adding the `--no-sandbox` flag for Chrome in Puppeteer. This disables some sandbox security in Chrome, but this is not relevant in our case since we're running this only on known `localhost` files _and_ only in a Docker container in a CI instance.


.circleci/config.yml
```diff
@@ -40,12 +41,12 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - npm-cache-{{ checksum "package-lock.json" }}
+            - npm-cache-{{ checksum "package.json" }}
```
thisisdano, line 44: For whatever reason, the package lock–related cache never seems to work

.circleci/config.yml
```diff
@@ -100,6 +102,7 @@ jobs:
           organization: uswds
   test_a11y_desktop:
     executor: my-executor
+    parallelism: 2
```
thisisdano, line 105: Added parallelism for fun

.circleci/config.yml
```diff
@@ -111,10 +114,9 @@ jobs:
       - run:
           name: Run pa11y-ci desktop
           command: npm run start-detached && npm run pa11y-ci:sitemap
-    # The resource_class feature allows configuring CPU and RAM resources for each job. Different resource classes are available for different executors. https://circleci.com/docs/2.0/configuration-reference/#resourceclass
-    resource_class: large
```
thisisdano, line 115: Moved this to the executor level

.pa11yci
```diff
@@ -2,8 +2,11 @@
   "defaults": {
     "userAgent": "pa11y",
     "timeout": 120000,
-    "concurrency": 2,
+    "concurrency": 10,
     "standard": "WCAG2AA",
+    "chromeLaunchConfig": {
+      "args": ["--disable-dev-shm-usage", "--no-sandbox", "--disable-gpu"]
+    },
```
thisisdano, line 9: Maybe this Chrome config for Puppeteer made a difference?

.pa11yci
```diff
@@ -2,8 +2,11 @@
   "defaults": {
     "userAgent": "pa11y",
     "timeout": 120000,
-    "concurrency": 2,
+    "concurrency": 10,
```
thisisdano, line 5: Not sure if this does much, but it did seem to do a tiny bit?

.circleci/config.yml
```diff
@@ -43,7 +44,7 @@ jobs:
             - npm-cache-{{ checksum "package-lock.json" }}
       - run:
           name: Install node dependencies
-          command: npm install
+          command: npm ci
```
thisisdano, line 47: Switched to `npm ci` (https://docs.npmjs.com/cli/v10/commands/npm-ci) so installing doesn't affect the package lock and we can take advantage of caching

.circleci/config.yml
```diff
@@ -5,6 +5,7 @@ orbs:
 
 executors:
   my-executor:
+    resource_class: large
```
thisisdano, line 8: Moved the resource class up here for simplicity

.pa11yci
thisisdano: Yeah, I guess I know what it's meant to do, but I'm not sure what the actual effect is in practice. I seems not to run 10x faster.

.circleci/config.yml
mejiaj, line 1: It's safe to remove the `Build USWDS components if needed` task [lines 69-71]. This task is also called on `npx gulp build`.

Tested in both [CircleCI](https://app.circleci.com/pipelines/github/uswds/uswds-site/6823/workflows/17bb2d18-2030-45b6-bcb4-c3dc5c326745/jobs/29731) and [Pages](https://pages.cloud.gov/sites/379/builds/375215/logs) without issues.

_It only runs once, so you'll see in logs it'll check and skip that task._

mejiaj (approved):
Nice work. I was able to verify that the tests pass.

At the end of the a11y tests there's this message that appears:
```bash
[2024-05-30 01:34:46] FATAL SignalException: SIGHUP
	/home/circleci/project/vendor/bundle/ruby/3.1.0/gems/webrick-1.8.1/lib/webrick/server.rb:173:in `select'
	/home/circleci/project/vendor/bundle/ruby/3.1.0/gems/webrick-1.8.1/lib/webrick/server.rb:173:in `block in start'
```
[Source → ](https://app.circleci.com/pipelines/github/uswds/uswds-site/6819/workflows/f5b72f49-0baa-4c8b-ba78-0b24c804fd1e/jobs/29716/parallel-runs/0/steps/0-104) 

Haven't been able to pinpoint the exact cause, but it doesn't look to affect the test.

PR #2805, opened 2024-09-03
USWDS-Site: Add changelog for Enable mixin in card component [#5974]

cathybaptista:
# Summary

Changelog for uswds/uswds#5974

## Related PR

uswds/uswds#5974

## Preview link

Preview link:
[Card changelog](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/cb-changelog-5974/components/card/#latest-updates)


mahoneycm (changes requested):
**praise:** I like that you linked the preview link directly to the changelog section! Thanks for that 😄 

**todo:** Small wording suggestion to add a bit of clarity to the changelog. Thanks @cathybaptista !

mahoneycm (approved):
LGTM! Updated the target branch to the `release-3.9.0` 👍

_data/changelogs/component-card.yml
```diff
@@ -2,6 +2,14 @@ title: Card
 type: component
 changelogURL:
 items:
+  - date: NNNN-NN-NN
+    summary: Added the `$theme-card-font-family` to the card container typeset mixin.
+    summaryAdditional: Now, cards respect the `$theme-card-font-family` theme setting.
```
amyleadem, line 7: **suggestion:** Recommend we re-write this to focus on the main thing the users will care about. 

```suggestion
    summary: Fixed a bug that caused the component to ignore the `$theme-card-font-family` setting.
```

_data/changelogs/component-card.yml
```diff
@@ -2,6 +2,13 @@ title: Card
 type: component
 changelogURL:
 items:
+  - date: NNNN-NN-NN
+    summary: Fixed a bug that caused the component to ignore the `$theme-card-font-family` setting.
+    affectsStyles: true
```
amyleadem, line 7: **question:** Do we want to mark this as a breaking change since this has the risk of changing the card font family? 
```suggestion
    isBreaking: true
    affectsStyles: true  
```

mahoneycm: I will say, it feels like the risk of change is rather low. Users would have to have previously set `$theme-card-font-family`, which would have previously not changed anything, and then keep the setting for it to have any "unexpected" settings. 

That said, happy to call this out if we're worried about it!

_data/changelogs/component-card.yml
```diff
@@ -2,6 +2,14 @@ title: Card
 type: component
 changelogURL:
 items:
+  - date: NNNN-NN-NN
+    summary: Fixed a bug that caused the component to ignore the `$theme-card-font-family` setting.
```
amyleadem, line 6: **suggestion (non-blocking)**: Since this is marked as breaking change, we should include  details about the break users should look for. 

```suggestion
    summary: Fixed a bug that caused the component to ignore the `$theme-card-font-family` setting.
    summaryAdditional: Confirm that your implementation of the card component displays with the expected font family. 
```

amyleadem (approved):

PR #2922, opened 2024-10-29
USWDS - Dependencies: Use node 20 for cg page builds

mejiaj:
# Summary

Node v23 is now LTS, but there isn't cg-pages support yet. Rolling this back to avoid build errors.

## Testing and review

If using NVMRC, install latest node 20 [`20.18.0`] version. 

```sh
nvm install lts/iron
```

All builds should pass.


mejiaj:
Closing since support has been added.

PR #3040, opened 2025-01-03
USWDS-Site: Add changelog for tooltip on non-button elements[#6035] 

amyleadem:
# Summary

Add a changelog for PR uswds/uswds#6035

>[!important]
> We need to update the changelog dates before merge.

## Preview link

[Tooltip changelog](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-changelog-6035/components/tooltip/#latest-updates)


_components/tooltip/guidance/usability.md
```diff
@@ -1,4 +1,3 @@
-- **Use affordances.** A hidden tooltip is unusable. Use tooltips only on elements that appear interactive, like buttons or links.
```
amyleadem, line 1: @jaclinec Wanted to check - Is it reasonable to remove this line now that we allow tooltip on non-interactive elements?

jaclinec (dismissed):
Yes, let's remove that line. Looks good!

_components/tooltip/guidance/usability.md
mahoneycm, line 1: **question:** Do we want to remove this completely? I thought we wanted to keep the description but update the lead. Let me know if I'm off here!

_data/changelogs/component-tooltip.yml
```diff
@@ -3,6 +3,13 @@ title: Tooltip
 type: component
 changelogURL:
 items:
+  - date: NNNN-NN-NN
+    summary: Enabled tooltip functionality on non-button elements.
```
mahoneycm, line 7: **suggestion:** Potential phrasing change to make the benefit a bit more clear:

```suggestion
    summary: Allow tooltip functionality on non-button elements.
    summaryAdditional: Now, tooltips can be applied to any element with the `usa-tooltip` class.
```

Let me know what you think!

_components/tooltip/guidance/usability.md
jaclinec: I'm not sure how one might use affordances on something like a static icon. But icons are supposed to convey meaning, so it's reasonable to think a user would want to hover or click on it to find out more info. I do think we should discourage using tooltips on things that don't seem interactive at all, like text on a page that doesn't have any other treatment to make it stand out.

Maybe this could be reworded to say something like: "Make tooltips discoverable. A hidden tooltip is unusable. Use tooltips on elements that appear interactive, like buttons, links, icons, or other elements that stand out." ?? tagging @finekatie for her input

finekatie: Hi. Wondering where this appears on the page. I can't see it in context. I think what @jaclinec has works, and I wonder about the use of "A hidden tooltip is unusable." Does it leave a user with more questions? Is that a necessary statement? If it's not needed, I'd remove it.

jaclinec: @finekatie The statement can be found on the USWDS tooltip page under "Usability guidance." I agree we can probably remove that sentence since it doesn't add anything.

So the statement would read: "**Make tooltips discoverable.** Use tooltips on elements that appear interactive, like buttons, links, icons, or other elements that stand out."

_data/changelogs/component-tooltip.yml
amyleadem: Ok I will leave as-is

_components/tooltip/guidance/usability.md
finekatie: I thought we were removing "A hidden tooltip is unusable" - the meeting notes didn't seem to address that - that I could tell...

amyleadem: It wasn't captured in the notes, but I recall that it was specifically only the bolded lead-in that we decided needed updating. I think we wanted to rest of it to stay as-is for now.

 Does it feel necessary to you to update?

finekatie: Only if @jaclinec has feelings about it...

jaclinec: I don't have strong feelings. It's fine to just change the bolded lead-in and leave the rest.

_data/changelogs/component-tooltip.yml
mahoneycm: Sounds good to me!

mahoneycm (approved):

PR #3042, opened 2025-01-05
Update Sass to modern syntax

thisisdano:
Description TK


css/_uswds-theme-custom-styles.scss
```diff
@@ -20,13 +16,6 @@
 // 3. Custom page styles
 // ----------------------------------------
 
-// TODO: move to uswds once finished
-$theme-site-cols-split: "tablet";
-$theme-site-cols-show: "desktop";
-$theme-grid-gap-default: "md";
-
-$site-measure: 5;
-
```
thisisdano, line 29: These are all now in variables so they can be `use`d properly

css/_uswds-theme-custom-styles.scss
```diff
@@ -112,10 +101,10 @@ abbr[title="required"] {
   $site-banner-icon-color: "blue-40v";
   $site-banner-action-color: "gray-20";
   @include typeset($site-banner-font-family);
+  background-color: color("ink");
```
thisisdano, line 104: They'll be a number of this type of change throughout, changing the order of rules to avoid splitting across nested rules...

css/_uswds-theme-custom-styles.scss
```diff
@@ -453,7 +442,6 @@ abbr[title="required"] {
 }
 
 .site-subheading {
-  @include typeset-h6;
```
thisisdano, line 456: Removed entirely, since moving to bottom caused some font issues. All it gets us is uppercasing, which I addressed in the previous PR

css/custom-styles/_card-policy.scss
```diff
@@ -1,3 +1,5 @@
+@use "uswds-core" as *;
```
thisisdano, line 1: You'll see this in all the files now...

css/custom-styles/_index.scss
thisisdano, line 1: Added an index for easier `forward`ing

css/custom-styles/_site-header.scss
```diff
@@ -41,7 +45,7 @@ $site-inner-nav-height: 6;
     }
 
     a {
-      color: color($site-ink);
+      color: color("ink");
```
thisisdano, line 48: Site ink is the same as ink

css/settings/_uswds-site-theme.scss
thisisdano, line 1: And lo! Only the settings we use! This took a bunch of diff'ing, but now we only have one small(ish) settings group instead of a bunch of redundant settings over 6 separate files

css/settings/old/_uswds-theme-color.scss
thisisdano, line 1: Moved all the old settings to a directory for reference. We can possibly just delete them now.

css/uswds-components.scss
```diff
@@ -1,9 +1,2 @@
...
-@import "./settings/uswds-theme-general";
-@import "./settings/uswds-theme-typography";
-@import "./settings/uswds-theme-spacing";
-@import "./settings/uswds-theme-color";
-@import "./settings/uswds-theme-utilities";
-@import "./settings/uswds-theme-components";
-
-@import "uswds/packages/uswds";
+@forward "settings/uswds-site-theme";
+@forward "uswds/packages/uswds";
```
thisisdano, line 2: Way simpler!

css/settings/_uswds-site-theme.scss
thisisdano: We can now reference this single file from all our entry points

package.json
```diff
@@ -43,10 +43,10 @@
     "prestart": "npx gulp build",
     "proof": "bundle exec htmlproofer --enforce-https=false --allow-missing-href=true --only_4xx --ignore-status-codes 403,429 --swap-urls 'https\\://designsystem.digital.gov/:/' --ignore-files '/whats-new/updates/(2017|2018|2019|2020|2021)/,/documentation/code-guidelines/' ./_site",
     "proof:all": "bundle exec htmlproofer --enforce-https=false --allow-missing-href=true --ignore-status-codes 0 --swap-urls 'https\\://designsystem.digital.gov/:/' --ignore-files '/whats-new/updates/(2017|2018|2019|2020|2021)/,/documentation/code-guidelines/' ./_site",
-    "serve": "npm run build:all-assets && bundle exec jekyll serve --drafts --future --incremental --livereload --host=localhost",
+    "serve": "npm run build:all-assets && bundle exec jekyll serve --drafts --future --incremental --host=localhost",
```
thisisdano, line 46: ymmv, but for me livereload just isn't worth it. I get pages and pages of `webrick` errors in the console, which is way too noisy to notice anything of value. The time saved not just hitting `command-R` is negligable

package.json
```diff
@@ -97,7 +97,7 @@
   },
   "snyk": true,
   "dependencies": {
-    "@uswds/compile": "1.2.1",
+    "@uswds/compile": "github:uswds/uswds-compile#dw-gulp-sass-update",
```
thisisdano, line 100: I'm using a feature branch here where I fixed the Sass JS API deprecation issues. We can link it to a formal compile release when that exists

thisisdano (commented):
Added code comments

css/_uswds-theme-custom-styles.scss
thisisdano: OK, to be fair, it DOES change the spacing a little between the eyebrow and the `h1`, but not enough to be an issue

mejiaj (dismissed):
LGTM, I didn't run into issues on compile or notice any regressions.

**Compile issues before/after.**
capture -Warp-2025-01-07@2x

package.json
mahoneycm: For what it's worth, I've pretty much exclusively used the local host link and not the live reload link in my time here ¯\_(ツ)_/¯

mahoneycm (dismissed):
Looks great! No perceived regressions and no build errors.

Really happy to have this update 👏

mejiaj (approved):
@thisisdano I've updated branch and resolved conflicts. This is ready for merge.

PR #3062, opened 2025-01-13
USWDS-Site: Add changelog for in-page nav `once` update [#5878]

amyleadem:
# Summary

Add a changelog for PR uswds/uswds#6172

>[!important]
> We need to update the changelog dates before merge.

## Preview link

In-page navigation changelog


mahoneycm (commented):
imo this doesn't feel like a change that warrants a changelog. I'd rather keep the changelog lean to things that affect usability or implementation.

That said, this changelog looks good and I'll approve if we want to move forward with this!

PR #3152, opened 2025-04-04
Update lifecycle-status.md

amycole501:
Added new proposal status for "Converting to pattern". Updated both mobile and desktop views.


pages/ui-components/lifecycle-status.md
```diff
@@ -15,6 +15,7 @@ changelog:
 {% assign component_status_items = site.data.lifecycle-status.components %}
 
 {% assign check_icon_classes = 'text-gray-60' %}
+{% assign converting_pattern_classes = proposal_classes %}
```
juliaelman, line 18: This should follow the same convention as the other classnames. Below is a suggestion for what to change this line to read as:

```suggestion
{% assign converting_pattern_classes = "lifecycle-bg--convert" %}
```

PR #3158, opened 2025-04-28
Ignore link checker on good first issue links

heymatthenry:
The links to the "good first issue" tag are a frequent cause of false positive HTML proofer errors, causing build failures and blocking site updates. This PR adds the `data-proofer-ignore` attribute to those links to prevent these false positives.


annepetersen (approved):
Links work, tests pass

PR #3215, opened 2026-06-03
USWDS-Site - Security: Fix highlighted nav item for security subpages

### LINKED ISSUE #2974: USWDS-Site: Fix "current" navigation on security updates
## Summary
The pages in the `security_updates` collection ([example security update page](https://designsystem.digital.gov/security_updates/santized-combo-box-content/)) do not show the correct "current" page in the side navigation. 

### Problem details
This is happening because [current.html](https://github.com/uswds/uswds-site/blob/main/_includes/nav/current.html#L7) connects the side nav with posts in a given collection by checking for the parent page's `subnav.type` and matching it with the collection declared in [config.yml](https://github.com/uswds/uswds-site/blob/main/_config.yml#L77). 

However, the [security.md](https://github.com/uswds/uswds-site/blob/main/pages/about/security.md?plain=1#L8) file has two `subnav` declarations in its front matter, which is causing `subnav.type` to be ignored. 

### Next steps

We should:
1. Consider if we still need these security updates from v2 on our site, or if they can be considered outdated. 
1. If we need them, consider replacing [`subnav.type: security_updates`](https://github.com/uswds/uswds-site/blob/main/pages/about/security.md?plain=1#L8) in the page's front matter with `type: security_updates`. 
    1. Note: This is dependent on the fix in PR #2958, which enabled connecting the side nav with a collection via the `type` data key. 

## Related issue
This is a follow up to the conversation in [PR #2958](https://github.com/uswds/uswds-site/pull/2958#discussion_r1842764668)

## Comment from #2958

> I recommend we remove the `subnav.type` check in lines 7-8 and replace it with the check added in lines 14-15.
> 
> Currently, there is only one other page that uses `subnav.type`: [security.md](https://github.com/uswds/uswds-site/blob/main/pages/about/security.md?plain=1#L8). Replacing `subnav.type` with `type` here should still connect it to the expected collection in [config.yml](https://github.com/uswds/uswds-site/blob/main/_config.yml#L77).
> 
> I don't see anything in the repo relating to `page.type` that would cause any regressions, but of course we would want to test that. If we move forward with this approach, we would also need to make an update to the front matter in security.md. 
> 
> Curious to hear thoughts on this approach and proposed additional change.

_Originally posted by @amyleadem in https://github.com/uswds/uswds-site/pull/2958#discussion_r1842764668_

olsonap:
# Summary

**Security subnav item hightlighted on Security subpages.** Updated the sidenav logic to properly highlight the Security sidenav item while on the security_updates pages.

## Related issue

Closes [#2974](https://github.com/uswds/uswds-site/issues/2974)

## Preview link

Preview link:
_Not available_

## Problem statement

The Security page provides links to two subpages -- Sanitized Combo Box content, and Automatic sanitizing for all JavaScript components.
As subpages of Security, the Security sidenav item should be highlighted (class="usa-current") while visiting these pages.
Instead, the sidenav has no highlighted items so users can easily get turned around.

## Solution

As suggested in [issue #2974](https://github.com/uswds/uswds-site/issues/2974), the solution was to remove `subnav.type: security_updates` in favor of `type: security_updates`.
Maybe it was in use once, but as of this PR there are no remnants of subnav.type remaining in the repo.
After implementing that change, it was noticed that the "Latest updates" subnav list item didn't do anything.
I've explicitly listed the path to `/about/security/#changelog` so users who click on it are navigated to the Security Latest Updates section.
An alternative would be to update the path of the Security subpages to exist under /about/security/updates/, and enable them to show as subnav items.

## Major changes

_Not applicable_

## Testing and review

Run this branch locally and navigate to /about/security/.
Click either of the "Security updates" links.
image

Note the highlighted sidenav item:
image
image

Compare to [the active site](https://designsystem.digital.gov/security_updates/santized-combo-box-content/) where no sidenav item is highlighted:
image

### DIFF
--- a/_includes/nav/current.html
+++ b/_includes/nav/current.html
@@ -4,8 +4,6 @@
   {% assign _current = true %}
 {% elsif include.primary and include.page.category == page.category %}
   {% assign _current = true %}
-{% elsif page.collection and include.page.subnav.type == page.collection %}
-  {% assign _current = true %}
 {% comment %}
   Use the page's `type` data key to connect the sidenav to a collection.
   This sets "current" for the pages inside the _posts directory.
--- a/pages/about/security.md
+++ b/pages/about/security.md
@@ -4,14 +4,13 @@ layout: styleguide
 title: Security
 category: About
 lead: Developing and using USWDS with security in mind
-subnav:
-  type: security_updates
+type: security_updates
 redirect_from:
 - /security/
 - /about/security-information/
 subnav:
 - text: Latest updates
-  href: '#changelog'
+  href: '/about/security/#changelog'
 changelog:
   key: about-security
 ---


ethangardner (approved):
LGTM! Thanks for the contribution.

PR #3217, opened 2026-06-04
USWDS-Site - Time Picker: Remove irrelevant copy and pasted guideline

### LINKED ISSUE #1398: Time Picker Page - Why Is There Information about District of Columbia on this Page
## Description

On the Time Picker page 9 https://designsystem.digital.gov/components/time-picker/ ), wondering why there is information about the District of Columbia on this page under the "Usability Guidance" section? I would expect the usability issues would be about inputting time into fields or using drop-down lists.

It might be because the section was copied from another page. Please check that the other items under this section make sense.

olsonap:
# Summary

**Removed guideline which doesn't apply to Time Picker.** The guideline removed applied broadly to select components, but was not relevant to the Time Picker component where the options are predetermined.

## Related issue

Closes [#1398](https://github.com/uswds/uswds-site/issues/1398)

## Preview link

_Not available_

## Problem statement

When the Time Picker component was added, some documentation was borrowed/copy and pasted from another component. The guideline removed in this PR is relevant for most selects, but not for this Time Picker because the options are predetermined.
This guideline should not appear as it makes it seem like the options are editable.

## Solution

Simply removed the guideline.

## Major changes

_Not applicable_

## Testing and review

I am running the site locally and it is rendering successfully.
image

### DIFF
--- a/_components/time-picker/guidance/usability.md
+++ b/_components/time-picker/guidance/usability.md
@@ -1,4 +1,3 @@
-- **Use option strings familiar to users.** The combo box filters by matching strings. Include option text that includes familiar strings or spellings (i.e. if using the combobox with a state list, include the postal abbreviation in the option text: District of Columbia (DC)).
 - **Make sure to test.** Test select menus thoroughly with members of your target audience. Several usability experts suggest they should be the “UI of last resort.” Many users find them confusing and difficult to use.
 - **Avoid dependent options.** Avoid making options in one select menu change based on the input to another. Users often don’t understand how choosing an item in one impacts another
 - **Use a good default.** When most users will (or should) pick a particular option, make it the default: `<option selected="selected">Default</option>`.


ethangardner (approved):
Thanks for this PR! This is approved, and I will leave the same comment I left for your other PR a moment ago in the sake of thoroughness.

**Just a heads up**: you may notice that the pages/build check for this PR shows as pending and never completes. I wanted to give you an update on what this means. 

**Why this happens**: for security reasons, our hosting provider, Cloud.gov Pages doesn't generate site previews for pull requests from forked repositories [source](https://docs.cloud.gov/pages/using-pages/previews/). Since this required check can't run on forked PRs, it will remain in a pending state.

**We're working on this**: We're discussing internally how to best handle this situation so community contributions aren't blocked by this required check. In the meantime, we'll review your changes and test them locally.

I'm going to add a label of "blocked" to this PR while we talk through our options. Know that no further action is required on your part. Thank you for your patience while we work through this process issue!

olsonap:
Thanks for the update, @ethangardner.

Issue #1388, opened 2021-12-17
On the Modal Component Page - An "id" in the Text is Not Marled as Code Item 


jfc3-dol:
## Description

On the Modal Component Page ( https://designsystem.digital.gov/components/modal/ ), the  "id" in the sentence below is not marled as a code item. Highlighted text in bold.

The **`id`** should belong to a paragraph or a brief piece of content.

The same applies for the following text.

must have a unique **`id`** so that

ryparker:
Closing as completed. Merged PR #3223 corrected the implementation-guidance occurrence. Both requested occurrences now format `id` as code: [unique IDs](https://github.com/uswds/uswds-site/blob/ffe2da7c59cf4e59881b1b0362a138440dd153ff/_components/modal/guidance/implementation.md) and [the paragraph/context guidance](https://github.com/uswds/uswds-site/blob/ffe2da7c59cf4e59881b1b0362a138440dd153ff/_components/modal/guidance/accessibility.md).

PR #3223, opened 2026-06-17
Format modal guidance id as code

arpitjain099:
This updates the modal implementation guidance so `id` is formatted as code in the "Use unique ids" sentence.\n\nFixes #1388.


ethangardner (approved):
Thanks for this PR!

**Just a heads up**: you may notice that the pages/build check for this PR shows as pending and never completes. I wanted to give you an update on what this means. 

**Why this happens**: for security reasons, our hosting provider, Cloud.gov Pages doesn't generate site previews for pull requests from forked repositories [source](https://docs.cloud.gov/pages/using-pages/previews/). Since this required check can't run on forked PRs, it will remain in a pending state.

**We're working on this**: We're discussing internally how to best handle this situation so community contributions aren't blocked by this required check. In the meantime, we'll review your changes and test them locally.

I'm going to add a label of "blocked" to this PR while we talk through our options. Know that no further action is required on your part. Thank you for your patience while we work through this process issue!

PR #3238, opened 2026-07-08
USWDS-Site - removing invalid emails from package.json

annepetersen:
Removing emails that no longer work for contributors who have left GSA


ethangardner (approved):
LGTM!

PR #3249, opened 2026-07-16
USWDS-Site - Site header: Name banner landmarks, drop redundant role

### LINKED ISSUE #2226: USWDS - Bug: Accessibility - Multiple unnamed banner landmarks
### Describe the bug

Both the `Banner` and `Header` components contain `<header>` elements, which show up in the rotor as banner landmark. It might not be clear to users with screen readers the difference between the two banner regions.

### Steps to reproduce the bug

1. Got to the USWDS homepage
2. Turn of VoiceOver
3. Pull up the rotor menu (VO+U)
4. Use the arrow keys to pull up the Landmarks
5. Notice two unnamed banners.

### Expected Behavior

One or or both of the components container a banner region should have an accessible name.

### Related code

_No response_

### Screenshots

Screenshot 2023-08-02 at 7 23 40 PM

### System setup

- Current USWDS
- MacOS
- VoiceOver

### Additional context

Some additional can be found in [Scott O'Hara's article on landmarks](https://www.scottohara.me/blog/2018/03/03/landmarks.html).

> Since there is no native “banner element” in HTML, modern browsers are meant to treat the `<header>`, that is the nearest direct child of the `<body>` element, as a banner landmark. While there can be multiple `<header>` elements used within a web document, ideally there should only be a single banner landmark exposed. **Instances where multiple banners may be desired would be if there were nested document or application roles within a single view. In such a case, it would be best to ensure each of these were given their own accessible name, as this would allow assistive technologies to jump to each nested banner and have a better idea of which “banner” they had navigated to.**

### Code of Conduct

- [X] I checked the [current issues](https://github.com/uswds/uswds/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22) for duplicate bug reports.

AKnassa:
# Summary

Cleaned up the site's banner landmarks for screen reader users: removed stray commas from aria-labels, dropped a redundant `role="banner"`, and gave the government banner and site header distinct accessible names.

## Related issue

Closes #2226

## Problem statement

Screen reader users hit two problems in the site's chrome: stray commas inside the banner and identifier aria-labels made VoiceOver stutter through the landmark names, and the site header carried a redundant `role="banner"` (core USWDS dropped that role back in v2.2.0), leaving multiple unnamed banner landmarks that were hard to tell apart.

## Solution

Three small edits, following the fix maintainers spelled out in the issue comments:

- `_includes/banner.html` — removed the six stray commas from the banner section's aria-label, and added `aria-label="U.S. Government"` to the banner header (the exact markup suggested in the comments).
- `_includes/navbar.html` — replaced `role="banner"` on the site header with an aria-label naming it after the website.
- `_data/identifier.yml` — removed the stray commas from the identifier's aria-label.

One thing intentionally left alone: the identifier *component preview examples* on the Identifier component page still show commas — that markup ships from the core `@uswds/uswds` package, not this repo, and maintainers asked for upstream changes to be tracked separately.

## Testing and review

- The site builds cleanly; the rendered pages show clean aria-labels, no `role="banner"` anywhere in the site chrome, and distinctly named banner and header landmarks.
- `bundle exec rspec` passes (10 examples, 0 failures).
- To review: open any page with VoiceOver or NVDA and cycle through landmarks — the banner and header now announce cleanly with distinct names.

### DIFF
--- a/_data/identifier.yml
+++ b/_data/identifier.yml
@@ -1,4 +1,4 @@
-ariaLabel: Agency identifier,,,,,,
+ariaLabel: Agency identifier
 logos:
   - path: img/gsa-logo.svg
     alt: GSA Logo
--- a/_includes/banner.html
+++ b/_includes/banner.html
@@ -1,9 +1,9 @@
 <section
   class="usa-banner site-banner"
-  aria-label="Official website of the United States government,,,,,,"
+  aria-label="Official website of the United States government"
 >
   <div class="usa-accordion">
-    <header class="usa-banner__header">
+    <header class="usa-banner__header" aria-label="U.S. Government">
       <div class="usa-banner__inner">
         <div class="grid-col-auto">
           <img
--- a/_includes/navbar.html
+++ b/_includes/navbar.html
@@ -1,6 +1,6 @@
 <a class="usa-skipnav" href="#main-content">Skip to main content</a>
 {% include banner.html %}
-<header class="usa-header usa-header--extended site-header site-header--dark" role="banner">
+<header class="usa-header usa-header--extended site-header site-header--dark" aria-label="{{ site.title }}">
   <div class="usa-navbar site-header__navbar">
     <div class="usa-logo site-logo" id="logo">
       <em class="usa-logo__text site-logo__text">


chandracarney:
Thanks for the contribution @AKnassa! I'm tagging in @jonathanbobel here for review.

PR #3253, opened 2026-07-17
Add USWDS init script to the site

ethangardner:
## Summary

Add `uswds-init` script to the site. This inlines the script because it's small and needed for first paint. Previously, the script was missing and causing the banner to briefly show as expanded before the JS executes.

## Preview link

Preview link: https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/task/add-uswds-init/

## Problem statement

The `uswds-init` script was missing and causing the banner to briefly show as expanded and abuptly collapse to it's expected state.

## Solution

Inline the `uswds-init` script in the `<head>` of the site.

## Testing
1. open the preview link and verify that the banner shows as collapsed. If you're working on a fast computer, it's helpful to [throttle your CPU to 4x or 6x slowdown using Chrome Devtools](https://www.debugbear.com/blog/cpu-throttling-in-chrome-devtools-and-lighthouse) to see behavior on the live site and verify that it does not happen on the preview link

PR #3259, opened 2026-07-19
USWDS-Site - Site header: Remove a broken leftover script tag

### LINKED ISSUE #2168: USWDS-Site - Bug: Duplicate uswds.min.js script
Script is loaded in head and scripts partial.

https://github.com/uswds/uswds-site/blob/2e6d8d642b6d26ce8560b596daf3604f9582cad5/_includes/head.html#L74
https://github.com/uswds/uswds-site/blob/2e6d8d642b6d26ce8560b596daf3604f9582cad5/_includes/scripts.html#L2

AKnassa:
# Summary

Removed a broken leftover script tag that made every page on the site request a directory as if it were a JavaScript file.

## Related issue

Closes #2168

## Problem statement

The issue reported `uswds.min.js` being loaded twice — once in the head and once in the scripts partial. Digging into the history shows the head copy was never actually removed: a later edit truncated the filename out of the tag instead of deleting it, leaving `<script src=".../assets/js/vendor/" type="text/javascript"></script>` — a script tag pointing at a directory — shipped in the head of every page since.

## Solution

Deleted that one broken line from `_includes/head.html`. The preload hint for `uswds.min.js` in the head and the real script load in the scripts partial stay exactly as they are — they pair correctly (same URL), so the page keeps its performance hint and its single, working script load.

## Testing and review

- Rendered pages verified before and after: the directory-request tag is gone, `uswds.min.js` appears exactly twice (one preload link, one script load) — checked on the homepage and a component page.
- `bundle exec rspec` passes (10 examples, 0 failures).
- To review: view source on any page of the preview build and search for `/assets/js/vendor/"` — no hits.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -81,4 +81,3 @@
 ================================================== -->
 <script src="{{ site.baseurl }}/assets/uswds/components/usa-banner.js.mjs" type="module"></script>
 <link rel="preload" href="{{ site.baseurl }}/assets/js/vendor/uswds.min.js" as="script">
-<script src="{{ site.baseurl }}/assets/js/vendor/" type="text/javascript"></script>


ethangardner:
Thanks for the contribution. This was also addressed in this PR which was recently merged: https://github.com/uswds/uswds-site/pull/3253. I'll close this one since the script tag has been removed.

PR #3282, opened 2026-08-05
USWDS-Site - Release 3.14.0 documentation

ethangardner:
# USWDS 3.14.0 documentation release

This PR consolidates all documentation updates for USWDS 3.14.0, including:
- Version reference updates (_config.yml, package.json)
- 37 changelog entries covering 29 PRs from [milestone #167](https://github.com/uswds/uswds/milestone/167)
- Component markup updates (breadcrumb variants, time-picker dependencies)
- Guidance updates (accordion, alert)

## Related USWDS release
- USWDS milestone: [uswds 3.14.0 - Aug 2026 (#167)](https://github.com/uswds/uswds/milestone/167)
- Release date: 2026-08-18
- [Preview link](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/release-3.14.0/)

### Version references (tasks 2.3, 2.5)
- `_config.yml`: `uswds_version: 3.13.0` → `3.14.0`
- `_config.yml`: hash URL updated to `uswds-3.14.0-zip-hash.txt`
- `package.json`: `@uswds/uswds` points to `github:uswds/uswds#develop`
  - ⚠️ **Action needed:** Revert to exact `3.14.0` once published to npm

### Component markup updates (task 2.6)
- **Breadcrumb**: Updated variant list to reflect wrap-by-default behavior
  - `.usa-breadcrumb--wrap` marked as default/no-op
  - `.usa-breadcrumb--truncate` added as opt-in single-line variant
- **Time picker**: Added `usa-combo-box` to dependencies

### Changelog entries (task 2.1)
37 entries covering 29 PRs across 20 changelog files:

| Component | PRs | Changes |
|---|---|---|
| Accordion | #6789 | Default accordion icon control placement update (potentially breaking) |
| Alert | #3288 | Guidance change for first word of copy |
| Banner | #6739, #6714 | JSDoc update, toggle lookup fix |
| Breadcrumb | #6722 | Wrap by default (breaking) |
| Character count | #6595, #6385, #6664 | aria-live defer, label selector, dynamic init |
| Combo box | #6736 | Input element overflow fix |
| Date picker | #6593, #6664 | aria-current, dynamic init |
| File input | #6669, #6660, #6664 | Error token, mobile text (breaking), dynamic init |
| Footer | #6674 | data-tag security hardening (potentially breaking) |
| In-page navigation | #6688 | data-enhanced standardization |
[...]

PR #3287, opened 2026-08-08
USWDS-site: test PR for CI debugging

ethangardner:


ethangardner:
I went down an absolute rabbit hole just to make sure there weren't pa11y regressions in #3282. 

The failing pa11y tests are a result of using a git SHA temporarily in the release branch as opposed to using a published release from NPM. The reason is the prebuilt package from npm is more strict about lockfiles and causing things to fail. Registry versions get normal npm cache/integrity behavior; git SHAs are re-fetched/rebuilt more eagerly. The pa11y-ci failures should resolve when we can use v3.14.0 from npm instead of a git hash.

PR #3288, opened 2026-08-12
USWDS-Site - Alert: Add new accessibility and usability guidance to ensure clarity

chandracarney:
# Summary
Add new accessibility and usability guidance to the alert component page to ensure clarity and consistency of information provided to end-users.

## Related issue
Closes [#6539](https://github.com/uswds/uswds/issues/6539)

## Problem statement
**All users should have clear understanding of the alert type** and the message that is being displayed on screen, including screen reader users and users with a range of disabilities.

**Screen reader users didn't have access to the type of alert that's conveyed to visual users** through the icon (e.g. warning, error, success), as it was unrecognized by screen readers due to the fact that it was programmatically decorative. 

## Solution
This is an update to the alert component's guidance. We aim to ensure teams are communicating the type of alert directly to  visual and non-visual users alike, so that users don't have to infer the alert level based on color and icon shape. This should also help teams use the correct type of alert for what they're communicating.

A potential solution would have been to include text alternative (alt text) for the icons, however, we ran into further considerations that steered us towards a guidance update:
* All users benefit from knowing the alert type presented (e.g. “error”), rather than depending on color and icon shape to infer information. This user group is broad and includes people who are color blind, have low vision, are distracted, have difficulty with pattern recognition, have memory issues, or are in bright/dim lighting.
* If we inject the error type in alt text, screen reader users might be getting duplicative information if a team has already put the alert type in the heading or body copy.
  * Depending on the inherent screen reader behavior, copy choice, and how the component is implemented, this duplicative information may come 3 times: “Alert, error alert, error, your password change was unsuccessful.”
* Web Component renderings cause things to break and behave in unexpected ways when injecting alt text to the icon in an alert, as reported by the VA Design System team using a Web Component alert. Their team highlighted that honesty & clarity to all users would be a better approach.
* The incorrect alert type is frequently used (e.g. warning alert used that should be an information alert); this would help design and content staff better choose the alert type.

## Testing and review
This is a guidance-only update, so you'll need to preview the site to view changes.


ethangardner:
> Web Component renderings of alert components cause things to break and behave in unexpected ways when injecting alt text to the icon in an alert.

Regarding the alt text and web component statement in the body of the PR, my gut tells me it would only be an issue if the web component used the Shadow DOM. If the component used the light DOM, the mechanics seem like they would  behave just like any other `img` with `alt` text. It would need some confirmation to see if I'm correct though.

chandracarney:
> Regarding the alt text and web component statement in the body of the PR, my gut tells me it would only be an issue if the web component used the Shadow DOM. If the component used the light DOM, the mechanics seem like they would behave just like any other `img` with `alt` text. It would need some confirmation to see if I'm correct though.

Totally, yep, this actually came from a VA finding when they were trying to inject alt text on their VADS Alert Web Component. I'll add to the main comment to clarify this was a verified finding.

annepetersen (approved):
LGTM! Very happy to get this out; I've heard it will be helpful for lots of teams.

annepetersen:
@chandracarney This commit needs a verified signature: could you rebase?

chandracarney:
@annepetersen done!

annepetersen (approved):
LGTM: thanks for the changelog addition

PR #3298, opened 2026-09-05
USWDS- Site: Add adaptive SVG favicon for full light/dark mode support

GovLandScout:
# Summary

_This PR added an adaptable SVG favicon that is compatible with both light and dark mode across sites. _

## Related issue

Closes #_3449_

_NOTE: I did not hear back from weather this should be distributed across all uses of USWDS, so this PR addresses the original request_

## Preview link

_N/A, this was tested with localhost_
Preview link:

## Problem statement

_The desired state is for the site's favicon to adapt to whether a visitor's browser is in light or dark mode. The actual state was a single dark-colored mark that works fine on a light tab bar but is very low-contrast/nearly invisible against a dark tab bar. This affects any visitor browsing designsystem.digital.gov with dark mode enabled at the OS or browser level._

## Solution

_This PR adds a new SVG favicon containing the existing USWDS logo as a vector path w/ an embedded <style> block using @media (prefers-color-scheme: dark) to switch the fill color between dark and black depending on user settings

This approach was chosen over two alternatives for these reasons;

Two separate PNG files + a media attribute on the <link> tag was tried first, but the media attribute isn't reliable so it did not actually switch in testing.
Runtime JS-based canvas inversion is technically possible but adds unnecessary complexity and a JS dependency for what can be a static asset.

The embedded-SVG-with-media-query technique (per the CSS-Tricks article referenced in the issue) proved reliable in local testing. The vector path was generated from the existing black mark via potrace and then class and <style> block were added to control fill color.

The new <link rel="icon" type="image/svg+xml"> tag was added to _includes/head.html, after the existing .ico and .png fallback links so that browsers without SVG favicon support continue to use the original static favicon, while modern browsers pick up the adaptive version._

## Major changes

_-Added img/favicons/favicon-adaptive.svg (new adaptive favicon)
-Modified _includes/head.html to add the new SVG favicon <link> tag alongside existing fallbacks._

## Testing and review

_1) Ran npm install and npm run serve to build and serve the site locally.
2)Verified the SVG's color-switching behavior directly (opening the SVG file alone) by toggling System Settings 
3) Appearance between Light and Dark and confirming the mark's fill color changed accordingly.
4) Verified the actual browser tab icon in Chrome (incognito, to avoid favicon caching) reflects the correct color in both modes.
At this stage, I'm mainly looking for confirmation on the scoping question raised above , and any feedback on browser/version compatibility the team wants me to additionally verify._


GovLandScout:
I did some double-checking, and I believe that the failed checks aren't caused by my PR work, and rather a div on /documentation/migration/

GovLandScout:
Is this a suitable fix for uswds/uswds#3449?

PR #3302, opened 2026-09-16
USWDS-Site - Code samples: Let long code lines wrap on small screens

### LINKED ISSUE #3293: Pa11y error thrown on accordions with specific conditions
## Summary
When accordions have elements that meet these conditions, pa11y and axe throw an error. The specific conditions are:
1. horizontally scrollable container (usually code blocks)
2. vertically scrollable container
3. small viewport
4. no focusable element

However, if a focusable element (like a link) exists, no error is thrown.

We suspect this is a false positive. However, we also discussed that this likely _isn't_ happening on larger viewports because a "copy code block" button is sometimes present, which means that a focusable element is present.

## Steps to reproduce
- Navigate to https://designsystem.digital.gov/documentation/migration/
- Narrow the browser viewport size to simulate a mobile experience (typically this will mean your browser viewport can't narrow any further)
- Increase your text size programmatically using `command` + `+` (or `ctrl` + `+`)
- Expand the first accordion with a heading "If you're using USWDS Gulp"
- Run axe browser extension
- Observe an error thrown for the first accordion element

Next:
- Open your browser's code inspector
- Under the first accordion element (with a heading "If you're using USWDS Gulp"), add an HTML link element
- Run axe browser extension again
- Observe no error

## Tasks
- [ ] Investigate
- [ ] Propose solution, which could include 
- ignoring specific pa11y tests upon build
- adding a copy code block button on smaller (mobile) viewports

AKnassa:
## Summary

Long paths and package names in code samples can overflow their containers on small screens. In the migration guide, that turns an accordion panel into a horizontally scrolling region without keyboard access.

Allow long code tokens to wrap, retaining `word-wrap: break-word` as a fallback. Add a mobile accessibility check that opens the affected accordion, and preserve that check when CI divides the sitemap across parallel jobs. Include the migration-page changelog entry.

Closes #3293.

## Screenshots

The same migration-guide code sample at a 375px viewport:

| Before | After |
| --- | --- |
| Long code tokens overflow the code sample | Long code tokens wrap within the code sample |

## Validation

Updated the branch from `main` at `922bbebe07678c205b21766a5211b76f64dd1530` and tested the combined result.

- Full asset/Jekyll build, lint, and 37 RSpec examples passed.
- Reproduced `scrollable-region-focusable` on `#m-a6` with the two new CSS declarations removed: a 335px panel contained 456px of content.
- With the fix, the expanded migration page passed axe at 375px, 320px, and 188px. The panel and its code examples no longer overflow.
- Enter opens the accordion and Space closes it. Copied code text is unchanged.
- The fallback passed at 375px and 188px with the newer `overflow-wrap` declaration removed.
- Verified one, two, and three CI shards: all 240 configured entries remain present, including the expanded-accordion check exactly once.
- Changelog formatting and the PR diff whitespace check passed.
- All nine reported CI checks passed on final commit `ca54fcaadcdfdd56b00d54367d609f3fd95cff1b`, including full mobile and desktop accessibility scans.

At 1280px, the local expanded-page scan in current Chrome reports three header color-contrast findings with both baseline CSS and the fix. The fix introduces no additional findings in that comparison.

## Review steps

Open `/documentation/migration/` at a narrow viewport, expand “If you're using USWDS Gulp,” and inspect the long `require(...)` examples. Confirm the complete code remains visible without sideways scrolling and the accordion remains keyboard operable.


ryparker (approved):
Reviewed the code-wrapping fix and the expanded-accordion regression check. No blocking findings.

Validated the final branch with current main merged in: full build, lint, and 37 RSpec examples pass. The baseline CSS reproduces the reported keyboard-scroll violation; the fix passes at 375px, 320px, and 188px. Keyboard open/close behavior and copied code text are preserved. The fallback also passes with the newer declaration removed, and the regression case is retained exactly once across one, two, and three CI shards.

The PR description records the unchanged desktop contrast findings separately and includes before/after screenshots.

Issue #2860, opened 2024-10-03
Migrate Data Visualization content into Pattern page


brunerae:
Migrate content from the Data Visualization Component page into a new Pattern page. As a first step, we'll need to plan for which sections of content from the Component page map to which sections on the Pattern page.

- [x] Create landscape analysis of other design systems' process of creating a pattern from a component

- [ ] Create [pattern proposal](https://docs.google.com/document/d/1oQPbQQDys1Sxe4iCDrHR4JtLblSLOjcGNBXx9grKu_g/edit?usp=sharing)

- [ ] Create a content outline that maps the [component page](https://designsystem.digital.gov/components/data-visualizations/) into a pattern page

- [ ] Create issue to build the page including how to link to the new pattern, meta data, page layout mockups and wireframes (if available)

amycole501:
Started landscape analysis of how other design systems transition a component to a pattern.

amycole501:
[Draft proposal started here](https://docs.google.com/document/d/1UUkuAD7GQh2kY4fqBP-jWDtvY1s80GwKLqYQhwe368A/edit?usp=sharing)

ryparker:
Thank you for the landscape analysis and draft planning recorded here. We are deferring the catalog migration and closing this implementation task as not planned, along with #3315 and uswds/uswds#6089. The existing guidance remains available; this closure does not claim the migration is complete.

The research and original comments are linked from [the data-visualization pattern discussion](https://github.com/uswds/uswds/discussions/6959). Please add concrete content or findability problems there if your team needs this change. That gives us one place to reconsider the direction and identify a focused implementation when appropriate.

PR #3315, opened 2026-09-18
docs(patterns): move data visualization guidance into patterns

ryparker:
Data visualization is guidance for choosing and presenting information, but the site currently lists it as a released component. Move that guidance to the Patterns section as **Visualize data**.

The existing chart examples, SVGs, data tables, section anchors, and update history remain intact. The former `/components/data-visualizations/` URL redirects to `/patterns/visualize-data/`. Navigation, the patterns overview, the table guidance link, and the component lifecycle inventory now point readers to the appropriate category.

Related to uswds/uswds#6089 and #2860. This implements the migration discussed in those issues; it does not introduce a chart library or change chart behavior.

Validation: rendered the pattern, overview, redirect, component index, and component status page with Jekyll 4.4.1 and cached repository data. Confirmed the redirect target, discovery links, retained chart SVGs and tables, section anchors, and removal from component listings. Reviewed the rename diff, which preserves 97% of the source. `git diff --check` passes. No application tests were needed for this documentation migration.


ryparker:
Closing this PR because the catalog migration is deferred. The research supporting a different classification remains useful, but changing navigation and URLs is not selected implementation work at this point. This PR does not fix chart behavior or add a chart component.

Continue the placement question in [the data-visualization pattern discussion](https://github.com/uswds/uswds/discussions/6959). It links the research in uswds/uswds#6089 and #2860, both of which are being closed as not planned. The patch and branch remain available. If the current placement causes a concrete problem for your team, please share the use case there so we can reconsider the appropriate scope.

